PR #47, opened 2026-07-12
Enable reduced-precision (Float16/Float32) runs of NonLinear_OneLayer_GML

michakraus:
## Motivation

The one-layer network variational integrator `NonLinear_OneLayer_GML` computed
several inner quantities in `Float64` regardless of the working precision, and
crashed at `Float16` while building the OGA dictionary. This prevented genuine
reduced-precision (`Float16`/`Float32`) integration — needed for a precision
sweep in [SolverBenchmark](https://github.com/michakraus/SolverBenchmark.jl).

## Changes (`src/network_integrators/NonLinear_OneLayer_GML.jl`)

- **`components!`** — evaluate the network at the interval boundaries with
  `[zero(ST)]` / `[one(ST)]` instead of the `Float64` literals `[0.0]` / `[1.0]`,
  so the residual and its derivatives are assembled genuinely at the working
  precision `ST`.
- **`initial_params!` (OGA1d)** — build the bias-dictionary range in `Float64`
  (`lo:step:hi` with `Float64` endpoints) so a large `dict_amount` no longer
  overflows a low-precision element type. Previously `Float16(400000) == Inf`
  gave a zero step (`ArgumentError: range step cannot be zero`). The greedy OGA
  fit remains in `Float64` — it is only a *seed* for the nonlinear solve — and is
  stored into the working-precision cache; the variational integrator equations
  and the Newton solve still run at precision `T`.
- **`initial_trajectory!` (IntegratorExtrapolation)** — use `[zero(h), h]` so the
  auxiliary sub-integration time axis matches the working precision.

## Tests

Adds `test/integrators/nvi_onelayer_reduced_precision_tests.jl`, which integrates
the harmonic-oscillator LODE at `Float64` and `Float32` and checks that the
solution type is genuinely `T` (no silent upcast) and tracks the analytic
solution. Bumps the version to `0.1.1`.

For `Float64` the OGA fit is numerically identical to before, so existing tests
are unaffected. The full suite passes locally.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "NonlinearIntegrators"
 uuid = "4ad56de4-0342-4c02-9694-ab2f23251d8c"
-version = "0.1.0"
+version = "0.1.1"
 authors = ["Michael Kraus <michael.kraus@ipp.mpg.de> and contributors"]
 
 [deps]
--- a/src/network_integrators/NonLinear_OneLayer_GML.jl
+++ b/src/network_integrators/NonLinear_OneLayer_GML.jl
@@ -244,7 +244,7 @@ function initial_trajectory!(sol, history, params, int::GeometricIntegrator{<:No
     local S = nbasis(method(int))
     local x = nlsolution(int)
 
-    tem_ode = similar(problem, [0.0, h], h / nstages, (q=StateVariable(sol.q[:]), p=StateVariable(sol.p[:])))
+    tem_ode = similar(problem, [zero(h), h], h / nstages, (q=StateVariable(sol.q[:]), p=StateVariable(sol.p[:])))
     tem_sol = integrate(tem_ode, integrator)
 
     for k in 1:D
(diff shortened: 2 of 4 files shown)


michakraus:
Addressed both review comments in fafeb22:

1. **`default_options` redefinition** — removed the override from the new test file. `regularization_factor` and `max_iterations` are now passed through `integrate(prob, method; …)` (verified that `integrate` forwards solver-option kwargs), so there is no longer a duplicate definition of `default_options` clashing with `nvi_onelayer_integrators_tests.jl` and no method-overwrite warning in the full run.
2. **`Float16` regression coverage** — added a `Float16` case with `dict_amount = 70000` (above `Float16`'s finite max ≈ 65504). It asserts the OGA dictionary range no longer throws `ArgumentError: range step cannot be zero`; the run now reaches the (still ill-conditioned) half-precision Newton solve, which the test tolerates. This locks in the range-construction fix without asserting a convergence that half precision cannot reach.

Full suite passes (9/9) with no overwrite warning.

PR #59, opened 2026-08-10
Refactor and build document

ZeyuanLee:
The main updates in this branch are:

- Introduced new types, `NetworkBasisCore` and `NetworkIntegratorCore`, to encapsulate common fields.
- Introduced `NetworkIntegratorCache` as a subtype of `IODEIntegratorCache`.
- Unified common functions in `NetworkIntegratorCore.jl.`
- Renamed some functions and arguments for improved clarity and consistency.
- Added necessary docstrings for types.
- Added more test sets covering all network integrators and their different field configurations.


michakraus (commented):
Thanks for this — the `NetworkIntegratorCore` / `NetworkBasisCore` split is the right
direction, and the docstrings are a real improvement. A few things need attention before
this can go in, starting with the CI failures.

## CI

**Documentation job — regression from this PR.** `docs/make.jl` regenerates the Benchmarks
figures by running the `quick` sweep, and in this run *every* case reported `FieldError`:

```
Float64 0.1  1/18  8  4 gelu  DogLeg  midpoint  16sqrt(eps)=2.4e-07 | FieldError — — 100 —
```

so no case produced data, all data-driven plots were skipped, and Documenter then failed
with 20 × `invalid local link/image: file does not exist` →
`makedocs encountered an error [:cross_references]`.

Cause: this PR removed the per-integrator `integrate!` that returned
`(sol = …, solving_time_list = …, …)` and replaced it with the unified version in
`NetworkIntegratorCore.jl` returning a plain `(sol, internal_values)` tuple, but
`benchmark/gml_benchmark_common.jl` still reads `res.sol.q` and `res.solving_time_list`.
The `solving_time` / `integrating_time` / `training_time` / `mse_err` cache fields are gone
too, so the `solve_secs` column has no source any more.

I pushed 6690ec7 to this branch, which adapts the benchmark harness (and
`oga_comparison.jl`, which had the same `res.sol` breakage), replaces `solve_secs` with
`total_secs` throughout the CSV/report/docs, and makes `docs/make.jl` fail immediately
with the list of missing figures instead of dying 35 minutes later in Documenter. Only
`benchmark/`, `docs/make.jl` and `docs/src/Benchmarks/benchmarks.md` are touched — no
`src/` changes.

One decision worth your input: per-step nonlinear-solve time is now unrecoverable. If you
want that metric back (it was more useful than wall-clock for solver comparisons), the
cleanest way is a `solving_time` field on `NetworkIntegratorCache` and a third return
value from the unified `integrate!`.

**Julia 1.10 (ubuntu + windows) and nightly — not your regressions.** Both fail identically
on `main`'s own push run (31228883474). 1.10 is the `Float16 OGA dictionary construction`
test: the ill-conditioned half-precision least-squares solve throws something other than
`SingularException` on Julia 1.10/x86_64 only (1.10-macOS is arm64 and passes, 1.12/1.13
pass everywhere). It should be resolved by the OGA rework on `feature/oga-variants`, which
replaces the LAPACK-only `qr(…, ColumnNorm())`/`svd` paths with generic fallbacks. Nightly
is an upstream precompilation failure in the Geometric* stack. Nothing for you to do on
either.

## Review

**1. The unified `initial_trajectory!` methods never run.** `NetworkIntegratorCore.jl`
defines generic versions for `IntegratorExtrapolation` (~line 90), `HermiteExtrapolation`
(~115) and `NoExtrapolation` (~156), but all five concrete types still carry
byte-identical copies (e.g. `NonLinear_OneLayer_GML.jl:194` vs the core version;
`Time_reversible_OneLayer.jl:182`; and `NonLinear_OneLayer_GML.jl:220` for
`NoExtrapolation`). The more specific method always wins, so the generic ones are dead
code and the duplication this PR set out to remove is still there. Deleting the per-type
copies — keeping only the ones that genuinely differ, e.g. the Hardcode/DenseNet Hermite
variants — makes the dedup real.

**2. `record_grid_points` is documented but unreachable.** `NetworkIntegratorCore` accepts
it and every cache uses `method.record_grid_points` to size `stage_values`, but no
integrator constructor accepts or forwards it — while three docstrings advertise it as a
user kwarg (`NonLinear_OneLayer_GML.jl:24`, `NonLinear_DenseNet_GML.jl:21`,
`Hardcode_int.jl:14`). Passing it throws a `MethodError`. Either forward it in all five
constructors or drop it from the docs. Same asymmetry for `show_status`: accepted by the
three Hardcode/time-reversible constructors, not by `NonLinear_OneLayer_GML` or
`NonLinear_DenseNet_GML`.

**3. `issymplectic` strengthened from `missing` to `true`.** This now claims symplecticity
for *all* `NetworkIntegratorMethod`s and the smoke tests assert it, where `main`
deliberately answered `missing` for `NonLinear_OneLayer_GML`. The scheme is variational,
but the per-step ansatz fit is approximate and neither the greedy/OGA seed nor the inexact
Newton solve is covered by a symplecticity proof. I'd keep `missing` unless there is a
result to cite — a trait that overstates a guarantee changes how downstream
GeometricIntegrators code treats these methods.

**4. `break` on NaN changes semantics silently.** The unified `integrate!` now breaks out of
the time loop on NaN, where `main` had `# break` commented out. Two consequences: the
remaining entries of `sol` keep their initialised values, so a run that blew up mid-way
returns a *finite-looking* final state — and the new tests assert only
`all(isfinite, collect(sol.q[:,1])[end])`, so they pass on a diverged run; and
`internal_values` is `Vector{Matrix}(undef, …)`, so entries past the break stay `#undef`
and throw `UndefRefError` when touched. Better to propagate a failure status than to
truncate quietly. (Pre-existing but adjacent: the loop writes `internal_values[n]`, which
is off by `n₁ - 1` whenever `n₁ > 1`.)

**5. The new tests are finiteness-only.** The added cross-products
(`onelayer_gml_unit.jl`, `hardcode_unit.jl`, `time_reversible_hardcode_unit.jl`,
`densenet_gml_unit.jl`, `dispatch_variants_unit.jl`) assert `assert_no_upcast` plus
`all(isfinite, …)` over a two-step run. Combined with (4) that is weak — a diverged or
truncated run passes. One accuracy assertion per dispatch variant, even a loose one, would
make them load-bearing.

**6. `test/unit/pr_integrator_unit.jl` is never included.** `test/runtests.jl` adds
`time_reversible_hardcode_unit.jl`, `densenet_gml_unit.jl` and `dispatch_variants_unit.jl`
only, so the new PR_Integrator test never runs.

**7. Stale comments after `70a0c1b`.** `densenet_gml_unit.jl` still claims the cross product
covers `HermiteExtrapolation`, but that commit removed Hermite from the lists — so the new
unified Hermite code paths are entirely untested while the comment says otherwise. Worth
fixing the comment and, ideally, keeping one Hermite case.

**8. `NetworkBasisCore` is type-unstable.** `activation` is the one field left untyped while
the other six are parameterised, so every access boxes and blocks specialisation in the
innermost basis evaluation. Add an `AT` parameter. It also has no docstring, unlike its
integrator counterpart.

**9. Version and dependencies.** The `integrate!` return shape and the
`initial_trajectory` → `initial_trajectory_method` rename are breaking, but
`Project.toml` still says `version = "0.2.0"` — needs a 0.3.0 bump (coordinate with
`feature/oga-variants`, which also bumps to 0.3.0). `BSplineKit` and `Infiltrator` were
correctly dropped from `src/NonlinearIntegrators.jl` but are still in `[deps]`/`[compat]`;
only `scripts/`, which has its own Project.toml, uses them.

**10. Minor.** `getproperty` forwarding has no matching `propertynames`, so tab-completion
and `propertynames(method)` no longer show the forwarded fields. `nbasis` is defined twice
for the same signature (once as `CompactBasisFunctions.nbasis`, once bare). The unified
`initial_guess!` reads the back-compat alias `method(int).initial_trajectory` rather than
the new canonical `initial_trajectory_method`. Dropping the old
`initial_trajectory!(…, ::Extrapolation)` fallback turns "extrapolation not implemented"
into a bare `MethodError`. `src/methods.jl` has no trailing newline, and
`obsolete/utilities.jl` adds 61 lines nothing includes.

## Merge order

`feature/oga-variants` rewrites the same integrator files (`Hardcode_int.jl` −108,
`NonLinear_OneLayer_GML.jl` −189, both `Time_reversible_*.jl`, `utilities.jl` −89) and
extracts the OGA into `src/oga/*`. Both branches independently added the `sol` argument to
`initial_params!`, but they will otherwise conflict heavily. Merging the OGA branch first
and rebasing this one on top is probably cheaper than the reverse, since this PR's changes
to those files are mostly mechanical renames.

PR #66, opened 2026-08-13
Build on GeometricIntegratorsBase alone, upgrade the Geometric* stack, retire GeometricMachineLearning

michakraus:
Upgrades to QuadratureRules 0.2 and CompactBasisFunctions 0.3, drops **three** dependencies that turned out not to be needed (GeometricIntegrators, GeometricMachineLearning, ContinuumArrays), and moves the training loops from GeometricMachineLearning to GeometricOptimizers.

## Dependency graph

Everything from the registry except GeometricOptimizers:

```
GeometricBase 0.14.8   GeometricEquations 0.21.2   GeometricIntegratorsBase 0.6.0
CompactBasisFunctions 0.3.1   QuadratureRules 0.2.1   SimpleSolvers 0.11.0
GeometricOptimizers 0.2.0  (git #main — registry still has only 0.1.0)
Zygote 0.7.12
```

Gone entirely: `GeometricIntegrators`, `GeometricMachineLearning`, `RungeKutta`, `GenericLinearAlgebra`, `ContinuumArrays` (still present indirectly via CompactBasisFunctions, as it should be).

## GeometricIntegrators is not needed

The package builds on `GeometricIntegratorsBase` alone.

- Every `GeometricIntegrators.Integrators.X` extension point was already a `GeometricIntegratorsBase` generic `import`ed into that module — the same objects — so ~40 call sites are simply requalified.
- `ImplicitMidpoint` comes from the base package, whose 0.6 release handles `IODEProblem`/`LODEProblem`. That is what the `IntegratorExtrapolation` warm start needs: it integrates a LODE sub-problem and reads `p` back out of it to seed the momentum degree of freedom. An ODE-only implicit midpoint fails with `FieldError: type NamedTuple has no field p`.
- `GeometricEquations` takes GeometricIntegrators' place in `[deps]`, because GeometricIntegrators was `@reexport`ing it: `AbstractProblemIODE` (18 uses), `StateVariable` and `initial_conditions` come from there, and GeometricIntegratorsBase only `using`s it. A swap, not an addition.
- `create_internal_stage_vector` was the only genuinely GeometricIntegrators-local name — three lines, now in `src/network_integrators/utilities.jl`.

Runge-Kutta reference integrators such as `Gauss(8)` are only needed by `benchmark/` and `scripts/`, which declare GeometricIntegrators in their own Project files.

**This also fixes Julia 1.13.** `GenericLinearAlgebra` overwrites a `LinearAlgebra` method, which 1.13 forbids during precompilation, and that took `RungeKutta` and hence `GeometricIntegrators` down with it. RungeKutta 0.6 had already dropped GenericLinearAlgebra; dropping GeometricIntegrators removes RungeKutta too, so neither is reachable. Verified loading on 1.13.0-rc2.

## GeometricOptimizers

The optimizer API has been retired from GeometricMachineLearning, so `Optimizer` / `AdamOptimizerWithDecay` / `GradientOptimizer` / `optimization_step!` are replaced by `Optimizer(ps, loss; algorithm, linesearch)` plus `solver_step!`/`update!`. `Adam` and `GradientMethod` supply only a direction, so the learning rate lives in the line search; `DecayingStatic(T; η₁=1e-3, η₂=5e-5, n=nepochs)` preserves the previous schedule exactly, decaying as `γ^t · η₁` with the same `γ = exp(log(η₂/η₁)/n)`.

The method and line search are now built at the parameter element type, where the old call passed Float64 constants regardless of the working precision.

Trained weights are **not** bit-identical to previous releases': the Adam bias-correction algebra differs, and gradients now come from the optimizer's own `GradientAutodiff` rather than explicit `Zygote.gradient` calls.

New `optimizer_params` / `network_params` (`src/network_integrators/utilities.jl`) convert between the nested per-layer parameters of `AbstractNeuralNetworks` and the flat `NamedTuple` of arrays GeometricOptimizers requires — its `OptimizerSolution` is an `AbstractVector`, a `Manifold`, or a flat `NamedTuple`, and network parameters are one level deeper. They alias rather than copy, so in-place updates stay visible through `PNN.params`.

## Other dependency audit results

- **GeometricMachineLearning** — after the optimizer move its only use was `NeuralNetwork`, which it `import`s straight from `AbstractNeuralNetworks`. The `_GML` type suffixes are kept.
- **ContinuumArrays** — no use in `src/`.
- **GeometricProblems** — now test-only. Its only `src/` use was `SINDy_methods/PR_Pretraining.jl`, never `include`d and calling into Flux (never a dependency); retired to `obsolete/script/`.
- Unused imports: `relative_maximum_error` (replaced by `GeometricSolution`, which the source actually names), `Options`, `NonlinearSolver`, `DogLeg`, and a no-op `using Base`.
- `Documenter` and `SharedArrays` were checked and already absent — Documenter lives only in `docs/Project.toml`, which is how CI's doctest job instantiates it. `Test` and `Logging` are correctly test-only.

## Reviewer notes

- **One `[sources]` git pin remains**, for GeometricOptimizers 0.2.0 (registry has 0.1.0), repeated in `benchmark/` and `scripts/` because `[sources]` does not propagate through a path dependency. A git `[sources]` entry blocks registration in General, so this cannot be tagged until GeometricOptimizers is released.
- **RungeKutta 0.6** from the original request is satisfied vacuously — RungeKutta is no longer a dependency of this package at any depth, so there is no compat entry to point at. It reaches `benchmark/` and `scripts/` via GeometricIntegrators 0.18.1.
- **Zygote compat widened to `0.6, 0.7`**, resolving to 0.7.12 — this is what [#41](https://github.com/JuliaGNI/NonlinearIntegrators.jl/pull/41) proposed, now closed as superseded. Zygote stays a direct dependency for `VNN_anstaz_zygote`, which supplies the velocity of the hardcoded ansatz in `Hardcode_int` and `Time_Reversible_Hardcode`.
- **Test status:** **1843/1843 pass** on the final stack — no GeometricIntegrators, GeometricIntegratorsBase 0.6.0, SimpleSolvers 0.11.0, Zygote 0.7.12, Julia 1.13.0-rc2 — in **8m19s**.
- **The performance concern is resolved.** An earlier intermediate stack ran the suite in 184 minutes, with a 3-epoch `TrainingMethod` integration burning 90 minutes of CPU. That was the fixed `Static(1e-3)` learning rate the older GeometricOptimizers forced: without the decay the loops never reached their `err < 5e-8` early exit and ran the full epoch budget every step. With `DecayingStatic` restored and GeometricOptimizers' Adam fixes in, the suite is back to 8m42s and no `∇F!` workaround is needed.


michakraus (commented):
Reviewed the full diff and the CI run for `100f81e`. The migration itself reads well — the requalification is mechanical and checkable, the `ImplicitMidpoint` reasoning is right, and the dependency audit is sound as far as it goes. Three defects, one of them a regression the PR description claims is resolved, plus a handful of smaller things.

## CI is not green, and the description says it is

The body reports **1843/1843 in 8m19s**. That number came from one local Julia 1.13 run. Against the CI matrix:

| job | this PR | `main` (run 31502231600) |
| --- | --- | --- |
| Julia 1.10 × 3 OSes | **fail in ~8s** (resolver) | pass |
| Documenter | **fail** (`Pkg` error) | pass |
| Julia 1.12 × 3 OSes | pass, **287m33s** of tests | pass, ~10m total |
| Julia ^1.13 × 3 OSes | pass, 19m57s of tests | pass, ~10m total |

The tests do pass everywhere they run. But two jobs never get to run them, and the 1.12 jobs take 4–5 hours against GitHub's 6-hour job limit.

---

## 1. `[sources]` without a `deps`/`extras` entry — breaks the Documenter job

`benchmark/Project.toml` and `scripts/Project.toml` list `GeometricOptimizers` under `[sources]` but in neither `[deps]` nor `[extras]`. Pkg rejects that outright. From the Documenter job, at the `julia --project=benchmark ... Pkg.instantiate()` step:

```
ERROR: Sources for `GeometricOptimizers` not listed in `deps` or `extras` section at
       ".../benchmark/Project.toml".
```

The comment above those entries — *"`[sources]` applies only to the active project, so the GeometricOptimizers pin … is not passed down through the path dependency and has to be repeated here"* — is also wrong, and `docs/Project.toml` was the evidence sitting in the same PR: it carries no such entry and the docs environment still resolved `GeometricOptimizers v0.2.0 #main`.

Verified both ways locally: with the entry removed, `julia --project=benchmark -e 'using Pkg; Pkg.develop(path=pwd()); Pkg.instantiate()'` (the workflow's exact command) and a bare `julia --project=scripts -e 'using Pkg; Pkg.instantiate()'` both resolve `GeometricOptimizers v0.2.0 #main`. The pin propagates out of the path dependency; repeating it was unnecessary as well as malformed.

**Fixed** by deleting the entry from both files.

## 2. Julia 1.10 cannot install the package

`[sources]` is a Pkg 1.11 feature. On 1.10 the table is ignored, so `GeometricOptimizers = "0.2"` has no candidate:

```
ERROR: Unsatisfiable requirements detected for package GeometricOptimizers [fc236c15]:
 └─restricted to versions 0.2 by NonlinearIntegrators [4ad56de4] — no versions left
```

`julia-buildpkg` fails in ~8 seconds on all three 1.10 jobs. This is a resolver limitation, not a source incompatibility, and it disappears when GeometricOptimizers 0.2 is registered and `[sources]` goes away — so `julia = "1.10"` stays as it is and the jobs stay red until then. **Documented** in the changelog's Known issues so the red jobs are not read as a code regression.

## 3. Julia 1.12 spends 287 minutes in type inference

This is the one I would not merge without a decision on. It is a regression: `main`'s 1.12 job is ~10 minutes.

Measured locally on 1.12.6 / 1.13.0-rc2, same machine, two-step harmonic-oscillator runs:

| | Julia 1.12 | Julia 1.13 |
| --- | --- | --- |
| `NonLinear_OneLayer_GML` + `OGA1d` (no optimizer) | 28.0 s | — |
| `NonLinear_OneLayer_GML` + `TrainingMethod` | **> 30 min** (killed) | 28.8 s |
| `NonLinear_DenseNet_GML` + `LSGD` | **> 7 min** (killed) | 7.8 s |

So it is the GeometricOptimizers-driven seeds specifically, and not `Adam` specifically — `LSGD` uses `GradientMethod` + `Static` and is affected too. Killing the process gives a stack in `jl_type_infer` → `typeinf` → `abstract_call`, and `--trace-compile` stops emitting new entries: inference, not numerical work. Consistent with the identical iterate counts and the fact that the tests do all pass.

Things I checked that it is **not**:

- **Not GeometricOptimizers [#35](https://github.com/JuliaGNI/GeometricOptimizers.jl/pull/35).** That fix is in the manifest (`Pkg.update` moves the pin from tree `48de9ef` to `75171ba`, and `rev = "main"` picks it up automatically) and the timings above are all *post*-#35. Its shape — `Optimizer` constructed and stepped in one inferred body — is the shape `initial_params!` has, but a standalone version of it now runs in **1.9 s on 1.12**, the same as on 1.13.
- **Not the loss evaluating a network under ForwardDiff.** Extending that repro so the loss runs a real `AbstractNeuralNetworks.Chain` on `NeuralNetworkParameters`, through the default `GradientAutodiff`: **1.97 s on 1.12**, 2.14 s on 1.13.
- **Not the parameter-conversion helpers**, though they were genuinely broken in a related way — see below. Making them inferrable does not move the number.

The blowup only appears once that optimizer call sits inside the `integrate` call graph. I have not isolated it further, and I do not think it is fixable from this repository without more work upstream. Left unresolved and **documented** in Known issues, with the measurements, so the next person does not start from zero.

## 4. `optimizer_params` / `network_params` are not inferrable

Found while chasing the above; a real defect on its own merits.

```julia
optimizer_params(ps) = merge(map(lname -> NamedTuple{map(f -> Symbol(lname, :_, f), ...)}(...), keys(ps))...)
```

The key set is built by `Symbol` concatenation at run time, which inference cannot fold:

```
julia> Base.return_types(NI.optimizer_params, (typeof(ps),))
1-element Vector{Any}:
 NamedTuple          # abstract
```

`@inferred` fails for both. That matters twice over: `ps_flat` reaches `Optimizer(ps_flat, loss; …)` with an unknown type, and `network_params` runs *inside the differentiated loss*, so it is on every function and gradient evaluation. **Fixed** by making both `@generated`; they now infer to the full concrete `@NamedTuple{L1_W::Matrix{Float64}, …}`, with the aliasing semantics unchanged.

They also had no direct test coverage, which is thin for the load-bearing piece of the migration — in particular nothing pinned the aliasing that `initial_params!` depends on when it reads trained weights back out of `PNN.params`. **Added** `test/unit/optimizer_params_unit.jl`, 22 assertions covering the round trip, key naming, the `NeuralNetworkParameters` method, the subset-of-layers case LSGD uses, element-type preservation, and aliasing.

---

## Smaller things (all fixed)

- **`Optimisers` is an unused dependency.** `src/NonlinearIntegrators.jl` has a bare `using Optimisers` with no call site anywhere in `src/`, `test/`, `benchmark/` or `scripts/`. Same reasoning that retired `ContinuumArrays` in this PR. Removing it drops it out of the manifest entirely, so nothing else in the graph wanted it either.
- **`src/methods.jl`** documents `TrainingMethod` as gradient descent "via `Optimisers.jl`" — it is `GeometricOptimizers.Adam` with a `DecayingStatic` line search now.
- **LSGD's `Static` line search is load-bearing.** #35 records under *Not fixed here* that `GradientMethod` with any *searching* line search throws `MethodError: no method matching gradient(::GradientCache)`, so even the default line search `Optimizer` would pick is unusable there. The loop only works because it passes `Static` explicitly. Added a comment so it does not get "simplified" away.
- **Stale prose the dependency removals invalidated**: `docs/src/index.md` and `VISE.md` still pointed at GeometricIntegrators.jl, `README.md` credited GeometricMachineLearning.jl for the network implementation. (`VISE.md`'s fences are unlabelled, so Documenter never executed them — accuracy, not a build failure.)
- **`Adam(PT)` / `GradientMethod()` were constructed twice** per loop, once for `Optimizer` and once for `OptimizerState`; bound to a local so the two provably share one method object.
- **`err = 0`** made `err` a `Union{Int,Float64}` in both DenseNet loops; `zero(PT)`.
- **Redundant `QuadratureRules.nnodes` qualifications** — the first commit message says these "can be unqualified whenever those files are next touched", and four of the ten sites are in files this PR touches. Now that `nnodes` comes from `GeometricBase` and QuadratureRules extends it, all ten are unqualified.

## Not changed

The `for i in 1:S₁` loops around the `x[...]` writes in `NonLinear_DenseNet_GML.jl` re-execute four loop-invariant assignments `S₁` times. Pre-existing and outside this diff.

michakraus:
Pushed the fixes as `1b1b484`, `9623788`, `3775aa5`.

**1. `[sources]`** — entry removed from `benchmark/Project.toml` and `scripts/Project.toml`. The Documenter job's *Install benchmark dependencies* step, which is where it failed, now passes.

**2. Julia 1.10** — left as it is and documented in the changelog's Known issues. `[sources]` is a Pkg 1.11 feature, so this is a resolver limitation rather than a source incompatibility, and it clears itself when GeometricOptimizers 0.2 registers and the section goes away. The three 1.10 jobs stay red until then.

**3. Julia 1.12** — **not fixed**, and I want to be clear that I could not root-cause it. What I established:

- It is the GeometricOptimizers-driven seeds specifically. Same integrator, same problem, two-step run: `OGA1d` 28.0 s on 1.12, `TrainingMethod` > 30 min on 1.12 and 28.8 s on 1.13. `LSGD` is affected too, so not `Adam`-specific.
- It is inference: the process sits in `jl_type_infer` → `typeinf` → `abstract_call` and `--trace-compile` stops emitting.
- It is **not** GeometricOptimizers #35, which is in the manifest — `rev = "main"` picks it up, and all the numbers above are post-#35.
- Standalone reproductions of the optimizer usage are all ~2 s on 1.12: construction and `solver_step!` in one inferred body; `Adam` + `DecayingStatic`; the loss evaluating a real `AbstractNeuralNetworks.Chain` on `NeuralNetworkParameters` through the default ForwardDiff gradient. The blowup only appears once that call sits inside the `integrate` call graph.

Recorded in Known issues with the measurements. Someone will need to take it further upstream; the 1.12 jobs will keep taking 4–5 hours in the meantime, against a 6-hour limit.

**4.** `optimizer_params`/`network_params` are `@generated` now and infer to concrete types; 22 new assertions in `test/unit/optimizer_params_unit.jl`. This does *not* move the 1.12 number — it is a separate defect that the search turned up.

Plus the smaller items: `Optimisers` dropped, the `TrainingMethod` docstring corrected, stale GeometricIntegrators/GeometricMachineLearning prose in `README.md`, `docs/src/index.md` and `VISE.md` updated, `Adam`/`GradientMethod` bound to a local so `Optimizer` and `OptimizerState` share one object, `err = zero(PT)`, a comment on why LSGD's `Static` is mandatory, and all ten `QuadratureRules.nnodes` sites unqualified.

Suite is **1865/1865 in 8m19s** locally on 1.13 (1843 + the 22 new), which is the same wall clock the description reports — so the 8m19s figure is right for 1.13 on that machine, it just is not what CI sees for 1.12. I could not get a like-for-like local `main` baseline, since `main`'s environment no longer resolves.

PR #73, opened 2026-08-14
Rename files and types to the GeometricIntegrators scheme

michakraus:
## Context

The layout and exported names dated from an exploratory phase. Directories were CamelCase (`src/CGVI_standard/`, `src/SINDy_methods/`) or split by implementation rather than method family (`src/network_basis/` + `src/network_integrators/` were two halves of one subsystem). Types carried underscores and dead provenance markers: `_GML` referred to GeometricMachineLearning, which stopped being a dependency earlier in this same release cycle; `Hardcode_int` named an implementation accident; `PR_Integrator`/`PR_Basis` named a paper where the documentation had always said VISE.

**The rename itself changes no numerics.** The one behavioural change in this PR is the `CGVINodal` `D > 1` fix below, which is the identity at `D = 1` and so leaves every existing result untouched.

## Directories

Source files now sit in one lowercase directory per method family. `network_basis/` and `network_integrators/` merge into `nvi/`, `CGVI_standard/` becomes `cgvi/`, `SINDy_methods/` becomes `vise/` — joining the already-clean `oga/`, which is unchanged. Tests, scripts, benchmarks and docs follow the same scheme.

Within a directory, each file is named after what it defines: `shallownet.jl`, `shallownet_basis.jl`, `densenet.jl`, `densenet_basis.jl`, `vise.jl`, `vise_basis.jl`, `cgvi.jl`. So the `Linear`/`NonLinear` and `_Int` markers disappear along with the type names that carried them, and `src/vise/` matches `src/cgvi/` in having its main file share the directory's name.

## Types

Integrators keep the bare family name and carry the variant as a suffix:

| old | new |
| --- | --- |
| `NonLinear_OneLayer_GML` | `ShallowNet` |
| `Hardcode_int` | `ShallowNetAutodiff` |
| `Time_reversible_OneLayer` | `ShallowNetReversible` |
| `Time_Reversible_Hardcode` | `ShallowNetAutodiffReversible` |
| `NonLinear_DenseNet_GML` | `DenseNet` |
| `CGVI_standard` | `CGVINodal` |
| `PR_Integrator` | `VISE` |

Bases take the `...Basis` names, so each matches the file that defines it, and the per-family abstract supertypes gain an `Abstract` prefix to make room:

| old | new |
| --- | --- |
| `OneLayerNetwork_GML` | `ShallowNetBasis` |
| `DenseNet_GML` | `DenseNetBasis` |
| `OneLayerNetBasis` (abstract) | `AbstractShallowNetBasis` |
| `DenseNetBasis` (abstract) | `AbstractDenseNetBasis` |
| `PR_Basis` | `VISEBasis` |

`OneLayerMethod` becomes `ShallowNetMethod`, and every `*Cache` follows its method. A call site now reads `ShallowNet(ShallowNetBasis{T}(tanh, 8), quad)`.

Note that `DenseNetBasis` is *reused*, not freed — it named the abstract supertype before and the concrete basis after — so the two renames had to be ordered.

## Three names that say something new

- **`Autodiff`** replaces `Hardcode`. The real distinction is the derivative backend: these two integrators differentiate a hand-written ansatz with `ForwardDiff`, where `ShallowNet` and `ShallowNetReversible` use derivatives compiled ahead of time by `SymbolicNeuralNetworks`.
- **`CGVINodal`** replaces `CGVI_standard`. It is *not* a copy of `GeometricIntegrators.CGVI`: upstream solves for all `S` basis coefficients plus the endpoint momentum (`D*(S+1)` unknowns), whereas this one pins `X[1] = q̄`, reads `q` back off the last coefficient and computes `p` explicitly, leaving `D*(S-1)`. That reduction requires an interpolatory basis with nodes at both endpoints — the coefficients *are* nodal values. The name also avoids a binding clash for anyone loading both packages, as `benchmark/` and `scripts/` do.
- **`VISE`** replaces `PR_Integrator`, matching what the documentation has always called it.

## Breaking

Old names are gone; there are no deprecation shims. Version stays at 0.3.0, which is unreleased.

`scripts/` JLD2 result keys are renamed too (`HO_PR_sol_q` → `HO_vise_sol_q`, and so on). The writer and both readers changed together, but **`.jld2` files produced before this change can no longer be read by the plotting scripts**. `scripts/results/` is git-ignored, so nothing in the repository is affected.

In `benchmark/`, `GML_BENCH_PRESET`/`SKIP_GML_BENCH` become `SHALLOWNET_BENCH_PRESET`/`SKIP_SHALLOWNET_BENCH`, and the combined-report figure prefix changes to `shallownet_benchmark`.

## Bug found and fixed

Diffing `CGVINodal` against upstream turned up a pre-existing bug: **it could not integrate a problem with more than one degree of freedom.** A `D = 2` run died in the first step with `SingularException: Zero pivot found at index 6`.

Four places index the flat vector of `S-1` free basis coefficients per degree of freedom, and two of them — `initial_guess!` (`x[D*(i-1)+k]`) and `residual!` (`b[D + D*(i-1)+k]`) — already agreed on *degree of freedom fastest, basis index slowest*. The other two did not:

- `components!` read `x[D*(d-1)+s]`, which is neither convention. At `D = 2, S = 4` that reads `x[3]` twice and never reads `x[6]`, leaving column 6 of the Jacobian identically zero. That is the zero pivot.
- `update!` assigned `sol.q .= nlsolution(int)[end]`, broadcasting one scalar across all `D` components of `q`.

Since the layout was in fact already settled by two of the four sites, there was nothing to decide and this is fixed here rather than deferred. Both corrections are the identity at `D = 1`, so the existing `D = 1` guards are unchanged to the last bit.

`test/unit/cgvi_unit.jl` gains a `D = 2` regression test built on `CoupledHarmonicOscillator` with the coupling parameter set to zero, which decouples it into two independent oscillators with different frequencies and different initial conditions — so each degree of freedom is checked against its own closed-form solution and a layout mistake cannot hide behind a merely worse number.

## Verification

- Full test suite: **1867/1867 pass**, both `Float64` and `Float32` (also re-run by the pre-push hook). The two new tests are the `CGVINodal` `D = 2` guard.
- The `D = 2` failure was reproduced against the pre-fix code (`Zero pivot found at index 6`) and both components now match their analytic references to `2.4e-13` (`q`) and `2.9e-12` (`p`).
- Full docs build succeeds end to end; figures regenerate under the new prefix and the site renders under `docs/build/{oga,vise,benchmarks}/`. The build caught one rename miss — a `@bibliography` block still pointing at `OGA.md`, which the case-insensitive macOS filesystem had masked — now fixed.
- `benchmark/run_harmonic_oscillator.jl quick` runs clean against the renamed harness and env var.


michakraus (commented):
## Review

I read the whole diff. **The rename itself is correct and complete.** Every type, cache, trait method, docstring, `include`, doc page, benchmark environment variable and JLD2 key was carried across; the `DenseNetBasis` abstract→concrete reuse is correctly ordered; the JLD2 writer and both readers agree on the new keys; `docs/make.jl`'s nav, figure prefixes and `@bibliography` `Pages` keys are consistent with the new paths; and no numerics changed. I also checked the places the rename deliberately did *not* touch — `obsolete/`, the root `README.md`, `scripts/test_CGVI_BSpline.jl`, `test_stability_analysis.jl`, `result_summary_table.jl`, `runtests.jl` — and none of them needed to change.

What is left is a handful of stragglers where the mechanical rename either missed a spot or applied the *type* map to something that was a *file path*, plus the pre-existing `CGVINodal` bug the PR documents. Details below; I am pushing fixes for all of them to this branch.

### 1. `CGVINodal` cannot integrate `D > 1` — the *Known issues* analysis is right

I confirmed this against the source and reproduced it. Worth adding: **there is no layout to choose.** Two of the four sites that touch the nonlinear solution vector already agree on *degree-of-freedom fastest, basis index slowest*:

- `initial_guess!` (`src/cgvi/cgvi.jl:166`) writes `x[D*(i-1)+k]`
- `residual!` (`src/cgvi/cgvi.jl:245`) writes `b[D + D*(i-1)+k]`

Only the other two disagree:

- `components!` (`src/cgvi/cgvi.jl:186`) reads `x[D*(d-1)+s]`, which is neither convention. At `D = 2, S = 4`: `d=1 → x[1],x[2],x[3]`, `d=2 → x[3],x[4],x[5]`. `x[3]` is read twice and `x[6]` never, so column 6 of the Jacobian is identically zero — that is the `Zero pivot found at index 6`.
- `update!` (`src/cgvi/cgvi.jl:270`) does `sol.q .= nlsolution(int)[end]`, broadcasting one scalar across all `D` components.

So the fix is two edits, both the identity at `D = 1`, which is the check that it is a fix rather than a change: the existing Float64/Float32 guards must stay bit-identical. Since the PR deferred it only because "fixing it means picking one layout, applying it in three places, and adding a `D > 1` regression test", and the layout is in fact already settled by two of the four sites, I have done it here rather than in a separate PR.

The regression test uses `CoupledHarmonicOscillator` with the coupling parameter `k = 0`, which decouples it into two independent harmonic oscillators with *different* frequencies and *different* initial conditions. Each component then has a closed-form reference, so a layout bug that duplicates or swaps components fails loudly instead of merely producing a worse number.

### 2. `src/vise/vise_integrator.jl:315` — stale file reference

```julia
# Argument order is (x, solver, args), as in `CGVINodal.jl` and the network
```

The original read `CGVI_standard.jl` — a *filename*, not a type. The type map was applied to it, so this now names a file that does not exist. Should be `cgvi/cgvi.jl`.

### 3. `scripts/parallel_run.sh:20` — the only executable line points at a path that has never existed

```sh
julia --project=script script/test_shallownet_reversible.jl $h $reg_factor &
```

The directory is `scripts/`, not `script/`. The PR updated the two commented-out sibling lines directly below to `scripts/…` and left the live one alone, so the file is now internally inconsistent. The `--project=script` is wrong for the same reason.

### 4. `src/vise/vise_basis.jl:44` — banner not renamed

The commented-out `show` method was renamed to `VISEBasis`, but its banner string still reads `======PR Basis by Symbolics======`.

### 5. Two source files whose name no longer matches their type

`src/nvi/shallownet_nonlinear.jl` defines `ShallowNet` and `src/nvi/densenet_nonlinear.jl` defines `DenseNet`. The `_nonlinear` suffix is the last surviving fragment of `NonLinear_OneLayer_GML`, and the only `Linear_*` counterpart is retired under `obsolete/`. These are the only two files in the PR that break the rule the description states for the bases ("each matches the file that defines it").

`src/vise/vise_integrator.jl` is a milder case of the same thing: `src/cgvi/cgvi.jl` sets the precedent that the directory name and the main file name agree.

Renamed to `shallownet.jl`, `densenet.jl`, `vise.jl` (and `test/unit/vise_integrator_unit.jl` → `vise_unit.jl`). Only four references exist and all are in `src/NonlinearIntegrators.jl` and `test/runtests.jl`.

### 6. `test/unit/` — the retired vocabulary survives in file-local identifiers

This is the only place in `test/` where it still appears:

| file | identifier | meaning |
| --- | --- | --- |
| `shallownet_unit.jl` | `build_ol_method`, `OL_INIT_METHODS`, `OL_EXTRAPOLATIONS` | OL = OneLayer |
| `shallownet_autodiff_unit.jl` | `HC_EXTRAPOLATIONS` | HC = Hardcode |
| `shallownet_autodiff_reversible_unit.jl` | `TRHC_EXTRAPOLATIONS` | TRHC = Time Reversible Hardcode |
| `shallownet_reversible_unit.jl` | `TROL_EXTRAPOLATIONS` | TROL = Time Reversible OneLayer |

Renamed to `build_shallownet_method` / `SHALLOWNET_INIT_METHODS` / `SHALLOWNET_EXTRAPOLATIONS` / `AUTODIFF_EXTRAPOLATIONS` / `AUTODIFF_REVERSIBLE_EXTRAPOLATIONS` / `REVERSIBLE_EXTRAPOLATIONS`. All test files are included into one module, so the new names have to stay distinct from each other and from the existing `DENSENET_*` constants — they do.

### 7. Two comments still say "hardcoded-ansatz integrator"

`test/unit/shallownet_autodiff_unit.jl:1` and `shallownet_autodiff_reversible_unit.jl:2`. This contradicts the PR's own reasoning that `Autodiff` replaces `Hardcode` because the real distinction is the derivative backend. Reworded to name the backend.

### 8. Pre-existing, unrelated: `docs/src/vise/vise.md` has no `#` heading

It opens at `## Basic Usage`, so Documenter renders "Basic Usage" as the page title. Added an H1 matching the `docs/make.jl` nav label while the file was being touched anyway.

PR #74, opened 2026-08-14
Update for SymbolicNeuralNetworks 0.4, benchmark the derivative backends

michakraus:
`SymbolicNeuralNetworks` 0.4.0 is a performance release with a source-compatible API: code generation performs common-subexpression elimination, batches are evaluated by an in-place kernel writing into one preallocated array, and an equation *set* (which is what `symbolic_pullback` returns) is generated as one function rather than one per leaf. Bumping the compat is the whole of the upgrade — the new `cse` and `inplace` keywords already default to the fast path, and no call site changed.

| | 0.3.7 | 0.4.0 |
|---|---|---|
| `ShallowNetBasis` build (relu³, S=4) | 0.95 s | 0.58 s |
| `ShallowNetBasis` build (tanh, S=8) | 0.42 s | 0.21 s |
| `DenseNetBasis` build (tanh, 3×3) | 3.22 s | 0.79 s |

The `ShallowNet` Float64 accuracy guard moves by 2 ulp (`0.38012229853795865` → `...837`), which is CSE reassociating arithmetic.

Two consequences are worth knowing and are in the CHANGELOG: the in-place result cannot be differentiated with `Zygote` (harmless here — the only `Zygote.gradient` in the package is on a hand-written ansatz), and the output element type is now promoted over the *inputs* rather than inferred from the expression, so reduced-precision networks round rather than widen.

## `symbolic = false`

Writing the benchmark surfaced that `ShallowNetBasis` compiled the symbolic derivatives even for the autodiff integrators, which never read them. `ShallowNetBasis{T}(σ, S; symbolic = false)` skips the build (0.095 s → 0.003 s at tanh/S=8); `ShallowNet`, `ShallowNetReversible` and `DenseNet` reject such a basis in their constructor rather than failing on a `nothing` call inside `components!`. `has_symbolic_derivatives` is exported.

Both bases also gain `cse` and `inplace`, forwarded to `build_nn_function` and defaulting to whatever that package defaults to. They exist to be turned *off*: `cse = false, inplace = false` is the code generation of 0.3.x.

## `benchmark/compare_derivative_backends.jl`

Compares `ShallowNet` / `ShallowNetReversible` (compiled derivatives) against `ShallowNetAutodiff` / `ShallowNetAutodiffReversible` (`ForwardDiff`), with the symbolic pair run under both codegen settings. Three measurements: the one-off basis build, the end-to-end solve split into a cold and a warm run, and the derivative kernels in isolation.

Per call, Float64/tanh:

| S | `cse+inplace` | `plain` | `ForwardDiff` |
|---|---|---|---|
| 4 | 0.083 µs / 528 B | 0.083 µs / 528 B | 1.38 µs / 6448 B |
| 8 | 0.084 µs / 720 B | 0.125 µs / 720 B | 2.96 µs / 16784 B |
| 16 | 0.166 µs / 1136 B | 0.792 µs / 2304 B | 7.92 µs / 54096 B |

The compiled kernels run 17–74× faster than `ForwardDiff` and allocate 12–68× less. Within the symbolic backend, 0.4.0's code generation is level with the old one at `S = 4` and 4.8–6.0× ahead at `S = 16`. The split is not the obvious one: for a *shallow* net `cse` buys nothing and generating the in-place kernel is the expensive half of the build (0.24 s vs 0.017 s), so `inplace` is what pays back at evaluation time; for `DenseNetBasis` it is the reverse, and `cse` is the entire 4.1×.

### What it does not measure

The report says both of these itself, rather than letting the tables imply otherwise:

- The two **backends** use different ansätze (raw network vs. boundary-interpolating) and different default OGA seeds, so accuracy and iteration counts compare *methods*; only the timings compare backends.
- The two **codegen** settings agree to machine epsilon at the kernel level (3e-17 at Float64) but not end to end. The residual stalls near the round-off floor, so a last-bit difference in the derivative decides which iterate Newton accepts — 3 of 8 paired `ShallowNet` cases stop after a different number of iterations, and `ref_err` moves by up to 200×. The report measures that amplification instead of asserting it away, and the unit test checks equality at the kernel level and finiteness only end to end.

## Open Issues

Four things noticed and not fixed are written up at the end of the CHANGELOG: the compiled kernels are still called once per quadrature node where 0.4 would allow one batched call; the autodiff pair computes the velocity with `Zygote` and its parameter gradient with `ForwardDiff` for the same expression; the residual floor above the convergence tolerance (two cases run the full 1000-iteration budget); and `default_iparams`, defined for three integrators and called nowhere.

## Verification

- Full suite: 1919 tests pass (+16 new), also via the pre-push hook.
- `benchmark/compare_derivative_backends.jl quick` runs clean from an empty results directory.
- `benchmark/run_harmonic_oscillator.jl quick` still produces its report (it shares `run_case` and the reporting helpers this touches).
- Docs build with no new warnings.


michakraus (commented):
Reviewed the whole diff. The upgrade itself holds up:

- The `shallownet_basis.jl` / `densenet_basis.jl` refactors are faithful to the code they replace — `symbolic_pullback(soutput, SNN)[1,1]`, `derivative(Jacobian(SNN))` and `symbolic_pullback(VNN[1,1], SNN)[1]` all reach `build_nn_function` with the same arguments as before, just routed through the `build` closure.
- The `symbolic = false` opt-out is safe: `ShallowNetAutodiff` and `ShallowNetAutodiffReversible` only ever touch `basis.NN` and `basis.activation` (`shallownet_autodiff.jl:256,273,453,458` and the reversible's `:262,279,462,467`), so the four `nothing`s are unreachable from them.
- `require_symbolic_derivatives` cannot be ambiguous — `NetworkBasis <: Basis` and both methods take `::AbstractString`, so the `NetworkBasis` one is strictly more specific.
- Pinning `SymbolicNeuralNetworks = "0.4"` rather than `"0.3, 0.4"` is right: `cse`/`inplace` do not exist on 0.3's `build_nn_function`.
- The new `==` bit-exactness assertion in `dispatch_variants_unit.jl` is not flaky. The seed path is deterministic (`Random.seed!(42)` in `initial_params!`, and the OGA dictionary is a `BiasGrid1d`, no RNG).

Everything below is a finding. Nothing is on the package's numerical path; the correctness issues are all in the new benchmark's `full` mode.

---

## Major — the report's row joins ignore `activation`, so `full` mode is wrong

`quick` has one activation, `full` has two (`tanh`, `gelu`). Two joins key only on `(problem, T, dt)`:

**`_codegen_endtoend_spread`, `compare_derivative_backends.jl:470`**

```julia
hit = findfirst(s -> s.method == base.label && s.problem == r.problem &&
                     s.T == r.T && s.dt == r.dt, rows)
```

In `full` mode a `gelu` plain-codegen row pairs against the `tanh` default-codegen row. The "max rel. Δ ref_err" and "cases with a different iteration count" columns then measure the activation difference — which is precisely the quantity this section exists to isolate from the codegen difference. Needs `s.activation == r.activation`.

**`_cell` (`:449`) and the head-to-head table (`:636-642`)**

Same key, so `findfirst` returns whichever activation comes first and the other one is silently absent — no row, no note, half the `full` sweep invisible. `compare_float16_activations.jl:114` gets this right because there activation *is* the compared axis; here it needs to be part of the row key and a table column.

The prose at `:542-544` ("Every other axis is held fixed: …") lists solver, initial guess, λ, steps and `R`/`S` but not activation, which is not in fact fixed in `full`.

## Moderate

**`run_kernel_benchmark` leaks the agreement handle on error (`:392`, `:433`).** `agr = open(agrpath, "w")` … `close(agr)` with the entire kernel sweep in between and no `try`/`finally`, right next to an `open(csvpath, "w") do io` that does it correctly. An exception in the timing loop leaves a truncated, unclosed `derivative_backends_codegen_agreement.csv` — and a later `DERIV_BENCH_REUSE=true` run reads it as complete.

**NaN counted as a disagreement (`:475`).** `d.iterations == r.iterations`, where both are `Float64` parsed from CSV and `NaN` when the solver state could not be read. `NaN == NaN` is `false`, so two equally-failed cases inflate the "different iteration count" column. `isequal` is the comparison wanted.

## Minor

**"Zero for the autodiff pair" is not zero** — `benchmark/README.md` and the generated report at `:596`. The `symbolic = false` build is timed with `@elapsed` like every other; `results/derivative_backends_quick.csv` records `basis_secs` of `5.8e-6`–`1.8e-3` for both autodiff rows. Small, but "zero" is a claim the file's own data contradicts.

**"five PNGs"** in `benchmark/README.md` — `write_backend_report` writes six (`_convergence_method`, `_convergence_heatmap`, `_runtime_vs_dt`, `_accuracy_vs_dt`, `_energy_drift_vs_dt`, `_iterations_vs_dt`).

**`using Random` is undeclared** (`:50`, for `MersenneTwister(42)` in `kernel_params`). It resolves only via the `@stdlib` entry of `LOAD_PATH`; `benchmark/Project.toml` lists `Dates`, `LinearAlgebra`, `Printf` and `Statistics` explicitly, so this one is an oversight rather than a convention.

**The CHANGELOG's own line references went stale in this commit.** `require_symbolic_derivatives` at `shallownet_reversible.jl:42` pushed the cited `185-186` to `186-187`; `densenet.jl:370` is a blank line, the bindings are `371-372`. They are also inconsistent — for `shallownet.jl` the citation `274-292` points at the per-node *loop*, for the other two at the local bindings. The loops are `shallownet_reversible.jl:217-240` and `densenet.jl:405-418`. (The rest check out: `shallownet_autodiff.jl:228`, `:230-232`, `:341`, `shallownet_autodiff_reversible.jl:351`, and all three `default_iparams` lines.)

**`cse`/`inplace` are pinned, not forwarded.** Three places — `shallownet_basis.jl:29-31`, `densenet_basis.jl` and `CHANGELOG.md:20-22` — say they "default to whatever `SymbolicNeuralNetworks` defaults to", but the signatures hardcode `cse::Bool = true, inplace::Bool = true`. The values do coincide with upstream (`build_function.jl:39`, `build_function_arrays.jl:29,87`) and hardcoding is the better choice — it keeps this package stable if upstream flips a default. It is the wording that should change, not the code.

**`NetworkBasisCore`'s docstring** (`network_basis_core.jl:4-21`) still says the derivatives "are built once at construction time and reused at every integration step". It is now the type that carries the four `nothing`s and should say so.

**`has_symbolic_derivatives`** (`network_basis.jl:49`) reaches through `b.common.dqdθ` although the `getproperty` forwarding two lines above exists so it does not have to.

**`plot_success_heatmap`'s local is still `strategies`** (`shallownet_report.jl`) now that `keyfn` is generic.

**`dispatch_variants_unit.jl`'s header** describes the file as covering initial-parameter-method dispatch variants; the two new testsets are about basis code generation.

**`DenseNetBasis`'s new `cse`/`inplace` have no test** — nothing constructs one with them.

---

The four items filed under CHANGELOG *Open Issues* (per-node kernel calls, the `Zygote`/`ForwardDiff` mismatch, the residual floor, the unused `default_iparams`) are pre-existing and I agree they belong in separate changes.

I am pushing fixes for everything above to this branch.

PR #75, opened 2026-08-14
Fix the anstaz → ansatz misspelling in the hardcoded-ansatz helpers

michakraus:
## What

The boundary ansatz of `ShallowNetAutodiff` / `ShallowNetAutodiffReversible` is built from nine small helpers whose names misspelled "ansatz" as "anstaz". Everywhere else in the package the word is spelled correctly (`_ansatz_modulation`, `_ansatz_target` in `src/oga/adapters.jl`, plus all prose), so these were the only holdouts.

| before | after |
|---|---|
| `NN_anstaz` | `NN_ansatz` |
| `VNN_anstaz`, `VNN_anstaz_zygote` | `VNN_ansatz`, `VNN_ansatz_zygote` |
| `∂NN_anstaz_∂params`, `∂VNN_anstaz_∂params` | `∂NN_ansatz_∂params`, `∂VNN_ansatz_∂params` |
| `∂NN_anstaz_∂q̄`, `∂NN_anstaz_∂q` | `∂NN_ansatz_∂q̄`, `∂NN_ansatz_∂q` |
| `∂VNN_anstaz_∂q̄`, `∂VNN_anstaz_∂q` | `∂VNN_ansatz_∂q̄`, `∂VNN_ansatz_∂q` |

Names only; no behaviour changed. None of them is exported, and the only call site outside `src/` was `benchmark/compare_derivative_backends.jl`, which reaches them through the `NI.` prefix. The identifier references in `benchmark/README.md` and in the CHANGELOG's `[Unreleased]` and `Open Issues` sections were updated with them.

## Also fixed: a `Float64` accumulator on the Newton path

Reviewing the rename drew attention to the places that write the four analytic boundary derivatives out by hand, and three of them carried `Float64` literals: `(1.0 - quad_nodes[j])` and `(-1.0)` in `src/nvi/shallownet_autodiff.jl:390-391`, `(-1.0)` in `src/nvi/shallownet_autodiff_reversible.jl:401`. They sit in the `p̄` row of `residual!`, where the accumulator `z` starts as `zero(ST)` and `ST` is a `ForwardDiff.Dual` during the Newton solve. Multiplying by a `Float64` literal promotes it, so at `Float32` the residual was summed as `Dual{…,Float64}` and only rounded back on the write into `b::Vector{ST}` — a silent upcast the suite cannot see, since `assert_no_upcast` checks the eltype of the final state, which is converted back, plus a mid-loop retype of `z` on the Newton path.

They are now integer literals, `(1 - quad_nodes[j])` and `(-1)`, which is the idiom `shallownet_autodiff_reversible.jl:400` already used: an integer literal takes the precision of its operand instead of imposing one. `Float64` results are bit-identical; `Float32` moves by round-off. Pre-existing, not introduced by the rename.

## Not fixed here

Two dead-code findings surfaced while doing it; both are recorded under `Open Issues` → *Dead code and documentation* instead:

- The four analytic boundary derivatives (`src/nvi/shallownet_autodiff.jl:234-238`) are defined and called nowhere. All four are written out by hand elsewhere: `residual!` spells the two `∂/∂q̄` ones into the `p̄` row (`shallownet_autodiff.jl:390-391`, `shallownet_autodiff_reversible.jl:400-401`) and `update!` spells the two `∂/∂q` ones into the momentum update (`:437-438`, `:446-447`, with the `∂VNN/∂q = 1` factor left implicit). Neither of those is `components!` and neither assembles a Jacobian — the solver differentiates one out of `residual!`. The Open Issues entry originally said otherwise and has been corrected.
- `src/nvi/shallownet_autodiff_reversible.jl:218-244` is a commented-out duplicate of the live definitions, already stale (`1.0 - t` where the live copy uses `one(t) - t`, so it predates the precision-generic refactor) and has to be hand-edited to stay in step — which this rename did, for a block no compiler checks.

## Verification

- `grep -rni anstaz .` matches only the three CHANGELOG lines that quote the old names in the new entry.
- `Pkg.test()` green, 1929/1929, both before and after the `residual!` literal fix.
- A `src/`-wide scan for bare float literals confirms those three were the only ones on a residual path.
- The benchmark script was not executed; its two references are `NI.`-qualified and resolve to the renamed functions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -252,6 +252,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **The `residual!` of the autodiff pair no longer accumulates in `Float64` at reduced
+  precision.** Three `Float64` literals sat in the `p̄` row of the residual — `(1.0 -
+  quad_nodes[j])` and `(-1.0)` in `src/nvi/shallownet_autodiff.jl:390-391`, `(-1.0)` in
+  `src/nvi/shallownet_autodiff_reversible.jl:401` — where they stand in for the analytic
+  boundary derivatives `∂q_h/∂q̄ = 1-t` and `∂v_h/∂q̄ = -1`. The accumulator `z` starts as
+  `zero(ST)`, and during the Newton solve `ST` is a `ForwardDiff.Dual`; multiplying by a
+  `Float64` literal promotes it, so at `Float32` the residual was summed as
+  `Dual{…,Float64}` and only rounded back on the write into `b::Vector{ST}`. That is a
+  silent upcast the suite cannot see — `assert_no_upcast` checks the eltype of the final
+  state, which is converted back — and it retypes `z` mid-loop on the Newton path. They are
+  now integer literals, `(1 - quad_nodes[j])` and `(-1)`, which is the idiom
+  `shallownet_autodiff_reversible.jl:400` already used and which takes the precision of the
+  operand instead of imposing one. `Float64` results are bit-identical; `Float32` moves by
+  round-off. Found while reviewing the `*_anstaz_*` → `*_ansatz_*` rename, which is what
+  drew attention to these hand-written copies of the boundary derivatives.
+
 - **`CGVINodal` (formerly `CGVI_standard`) can integrate problems with more than one degree of
   freedom.** Found while diffing it against `GeometricIntegrators.CGVI` for the rename above, and
   pre-existing rather than introduced by it: a `D = 2` run (Lagrange basis on 4 Lobatto nodes)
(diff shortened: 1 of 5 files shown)


michakraus (commented):
## Review

The rename is correct and complete. What I checked:

- `grep -rni "anstaz" .` matches only the three CHANGELOG lines that quote the old names deliberately. Nothing left in `src/`, `test/`, `docs/`, `scripts/` or `benchmark/`.
- Nothing in `test/`, `docs/` or `scripts/` ever referenced any of the nine names, so the claim that `benchmark/compare_derivative_backends.jl` was the only external call site holds.
- Both files are `include`d into the same module, so the reversible integrator resolves the renamed definitions from `shallownet_autodiff.jl:223-238`. The `NI.`-qualified benchmark references were updated with them.
- All nine names are unexported, so this is not a breaking change and `### Changed` is the right home for it.

Three things came out of the review. One is a wrong statement in the text this PR adds, one is a real defect next door to it, and one is a note on the scope decision.

### 1. The new Open Issues entry names the wrong function — fixed

`CHANGELOG.md` said of the four analytic boundary derivatives:

> `components!` writes those same four quantities out by hand where it assembles the Jacobian.

Both halves are wrong, and it matters because the entry's remedy is "either call them there or drop them" — so it points a future reader at the wrong call sites.

`components!` writes no such thing; it only fills the parameter derivatives `dqdW2c` / `dvdW1c` / … from `∂NN_ansatz_∂params` and `∂VNN_ansatz_∂params`. The hand-written copies are split across two other functions, and they assemble the **residual**, not a Jacobian — the solver differentiates the Jacobian out of `residual!`:

- `∂NN_ansatz_∂q̄ = 1-t` and `∂VNN_ansatz_∂q̄ = -1` are in `residual!`, in the `p̄` row: `shallownet_autodiff.jl:390-391`, `shallownet_autodiff_reversible.jl:400-401`.
- `∂NN_ansatz_∂q = t` and `∂VNN_ansatz_∂q = 1` are in `update!`, in the momentum update: `shallownet_autodiff.jl:437-438`, `shallownet_autodiff_reversible.jl:446-447`. The `1` is implicit there, as a bare `z += b[j] * P[j][k]`.

Corrected in the CHANGELOG and in the PR description.

### 2. Those hand-written copies carried `Float64` literals — fixed

Pre-existing, not introduced here, but it is exactly the code the new entry points at:

```julia
z += timestep(int) * method(int).b[j] * F[j][k] * (1.0 - quad_nodes[j])   # shallownet_autodiff.jl:390
z += method(int).b[j] * P[j][k] * (-1.0)                                  # :391, and _reversible.jl:401
```

`z` starts as `zero(ST)`, and during the Newton solve `ST` is a `ForwardDiff.Dual`. Multiplying by a `Float64` literal promotes it, so at `Float32` the `p̄` row of the residual is accumulated as `Dual{…,Float64}` and only rounded back on the write into `b::Vector{ST}`. Two consequences:

- A silent upcast the suite cannot see. `assert_no_upcast` (`test/testsetup.jl:73`) checks the eltype of the final state, which is converted back on assignment — so the gate passes while the arithmetic in between runs at double precision.
- A mid-loop retype of `z` on the Newton hot path.

`shallownet_autodiff_reversible.jl:400` already had the right idiom on the adjacent line — `(1-quad_nodes[j])` with an integer literal, which takes the precision of its operand rather than imposing one — so the fix is to make the other three match it. A `src/`-wide scan for bare float literals confirms these three were the only ones on a residual path.

`Float64` results are bit-identical (`1 - x === 1.0 - x` for `x::Float64`); only `Float32` intermediates move, by round-off, and the `Float32` guards are tolerance-based (`err < 1e-3`) rather than hardcoded digits. Full suite green after the change.

### 3. On the two items left under *Not fixed here*

Both stay, deliberately, but worth stating why the second one is the more expensive of the two to defer.

The four dead helpers are genuinely awkward to revive: their six-argument signature discards five arguments, so calling them from `residual!` / `update!` for what is `1 - t` reads worse than the literal it would replace. Leaving them documented is defensible; if they are ever picked up again, the signature is the thing to fix first.

The commented-out block at `shallownet_autodiff_reversible.jl:218-244` is the one with a running cost. This PR is the demonstration: nine identifiers had to be renamed inside it by hand, in a block no compiler, test or linter checks, and it still disagrees with the live copy on `1.0 - t` vs `one(t) - t`. Every future edit to the live definitions pays the same tax or lets the drift widen. The Open Issues entry is accurate that it should just be deleted.

---

The rename is sound. GitHub will not let me formally approve my own PR, so this is a comment review; the two fixes above are on a follow-up commit to this branch.

PR #76, opened 2026-08-14
Update for SymbolicNeuralNetworks 0.5

michakraus:
[SymbolicNeuralNetworks 0.5.0](https://github.com/JuliaGNI/SymbolicNeuralNetworks.jl/blob/main/CHANGELOG.md) is a refactor of the whole package below its exported surface, with no deprecation shims. Exactly one of its breaking changes reaches this package.

## The breaking change

`symbolic_pullback` is now `symbolic_parameter_gradient`, and for a **scalar** expression it returns the parameter-shaped gradient directly instead of a one-element array holding it. Both bases were differentiating a scalar and unwrapping the array afterwards, so the unwrapping goes with it — and it had to: indexing a `NeuralNetworkParameters` yields its first leaf, not the whole set, so leaving the `[1]` in place would have silently dropped `L2` from `dvdθ`.

## What does *not* change

| 0.5 breaking change | impact here |
|---|---|
| `nn.input` is a `Vector{Num}`, not a `Symbolics.Arr` | none — only ever passed through, and `build_nn_function` accepts either |
| `build_nn_function` returns a struct, not a closure | none in source; `NetworkBasisCore`'s `QWFT`/`VT`/`VWFT` parameters just become concrete and inferable |
| `Jacobian(f, nn)` flattens a non-vector `f` with `vec` | none — `f` is a length-1 vector here, so the 1×1 result is unchanged |
| `input_dimension`/`output_dimension` come from `AbstractNeuralNetworks ≥ 0.6.4` | compat bump only; those names are unused here and 0.6.4 is additive over 0.6.3 |
| generated kernels reserve `out`, `x1`, `x2`, `ps`, `k` | none — no free symbolic variables in our equations |

## Also in here

`DenseNetBasis` now builds `dvdθ` the way `ShallowNetBasis` does, from the scalar entry of the 1×1 Jacobian rather than from the whole array. The two had drifted apart, and under 0.5 the array form wraps a one-element array of equation-set functions in an indirection for nothing. `DVDθ` in `DenseNet.components!` consequently returns a parameter set like `DQDθ` does, and the `[1,1]` that used to unwrap it is gone.

The benchmark and the CHANGELOG framed `cse = false, inplace = false` as "the code generation of 0.3.x" and the defaults as "the 0.4.0 defaults". 0.5 rewrote code generation on the syntax tree, so that attribution no longer means anything; both now describe what the two keywords do.

## Verification

- **Kernels against `ForwardDiff`**, off the integrator: `dqdθ`, `dvdθ` and `V_func` agree to ≤ 5.6e-17 at Float64 for both bases, with the full `L1`/`L2`(/`L3`) nesting and densenet's `dvdθ` flattening to exactly `NP`.
- **Test suite**: 1929/1929, at Float64 and Float32, including the `assert_no_upcast` guards against a silent precision change through 0.5's `promoted_eltype` allocation path.
- **`benchmark/compare_derivative_backends.jl` re-run** in `quick` mode against 0.5. The two codegen settings still agree to round-off (≤ 8e-17 at Float64, ≤ 8e-8 at Float32), so the timings are trustworthy: compiled kernels 18–62× faster than `ForwardDiff`, allocating 12–68× less. The refreshed kernel table replaces the stale one in the CHANGELOG.
- **Docs build** clean, doctests included.

Two docstring build times had drifted far enough to be wrong and were re-measured warm: the `DenseNetBasis` `cse = false` build is 110 ms against 27 ms (still ~4×, but the old 3.22 s / 0.79 s figures were an order of magnitude out), and the `symbolic = false` shallow build is 29 ns against 15 ms. Both now say they are warm measurements — the first build in a fresh process is ~9.3 s either way, which is compiling the code generation rather than running it.


michakraus (commented):
## Review

The core change is right, and I checked it rather than taking the description's word for it: the
0.5 semantics are what the PR says they are (`symbolic_parameter_gradient` dispatches on
`AbstractArray`, so a scalar `f` goes to `_parameter_gradient(f, differentials)` and comes back
as a bare parameter set — `SymbolicNeuralNetworks/src/derivatives/gradient.jl:104-110`), and the
compiled kernels agree with an independent `ForwardDiff` reference taken off the integrator:

| basis | `dqdθ` | `dvdθ` | `V_func` | flattened length |
|---|---|---|---|---|
| `ShallowNetBasis{Float64}(tanh, 8)` | 7.4e-17 | 3.8e-17 | 1.5e-16 | 24 = 3·S |
| `DenseNetBasis{Float64}(tanh, 3, 3)` | 2.4e-16 | 7.2e-16 | 3.3e-16 | 21 = `NP` |
| both, `Float32` | ≤ 5.3e-8 | ≤ 9.4e-8 | ≤ 8.6e-8 | — |

(max relative deviation, each under both codegen settings). The docstring build times reproduce
too: warm `ShallowNetBasis(tanh, 8)` 14.2 ms, `DenseNetBasis(3,3)` 26.1 ms with the defaults and
106.4 ms without, first build in a fresh process 8.9 s. So the re-measurement the PR did was
worth doing and the new figures stand.

No defects in the code. Six things worth fixing, all now pushed to the branch.

### 1. `benchmark/README.md` is the one file the 0.4.0 → 0.5 rewrite missed

The PR retires the "this is what 0.4.0 bought" framing in the script header, in
`write_backend_report` (`benchmark/compare_derivative_backends.jl:553-557`), in the CHANGELOG and
in both basis docstrings — but `benchmark/README.md:96-102` still carries the old table verbatim
("what `SymbolicNeuralNetworks` 0.4.0 changed", "the 0.4.0 defaults", "`plain`: the code
generation of 0.3.x"). That is the file CHANGELOG.md:58 sends the reader to. Its agreement figure
at line 131 (3e-17) is also the last uncorrected copy of the number the PR moved to ≤ 8e-17
everywhere else; the re-run's `derivative_backends_codegen_agreement.csv` maxes at 7.6e-17.

Fixed: the README now mirrors the wording the report generator emits.

### 2. The PR makes the two derivative-build blocks identical, then comments instead of calling

After this change `src/nvi/densenet_basis.jl:43-54` and `src/nvi/shallownet_basis.jl:85-98` are
line-for-line the same modulo which `NN` they are handed — and the PR notices, adding "As in
`build_shallownet_derivatives`" rather than a call. But drift is exactly what a comment cannot
prevent, and the drift being repaired here is the reason the entry exists.

Fixed: `build_shallownet_derivatives` → `build_network_derivatives`, moved next to
`NetworkBasisCore` (whose four symbolic slots it fills), called by both bases. Unexported, no
`@docs` entry; the only reference outside `src/` was a comment in the benchmark, updated.

### 3. Leftover vocabulary from before the rename

- `src/nvi/shallownet_basis.jl:76-78` — the builder's docstring still described its job as "the
  pullback of the output with respect to the parameters … and the pullback of *that*".
  `symbolic_pullback` no longer exists upstream, and every other mention in the PR moved to
  gradient wording. Rewritten with the function's new home.
- `src/nvi/shallownet_basis.jl:46` — the example binds `legacy_codegen = ShallowNetBasis{Float64}(tanh, 8; cse = false, inplace = false)`.
  "Legacy" *is* the 0.3.x framing retired three lines above. Renamed to `plain_codegen`, which is
  what the benchmark and the CHANGELOG table call it.

### 4. Two of the new Open Issues entries are wrapped mid-clause

`CHANGELOG.md:846-847` and `870-872` break at 40 and 10 columns against the file's 92–95
("but it / dominates", "now are, / which would make it return a scalar"). Re-wrapped. Also
normalised `≤8e-17` against the `≤ 5.6e-17` spelling two paragraphs away.

### 5. The `V_func` entry contradicts itself

`CHANGELOG.md:866-869` says "every call site immediately indexes `[1]`" and then "the single call
site that consumes it", and misses `test/unit/dispatch_variants_unit.jl`, which unwraps it too.
Reworded. I also confirmed the fix the entry proposes would work — `build_nn_function` accepts a
scalar expression and returns a plain `Float64` — and noted that in the entry, but left the
change itself out: it is a shape change to the public `basis.V_func` field and it stays recorded
as an Open Issue.

### 6. Closed the first Open Issue instead of only recording it

The entry is right that this is the gap that mattered here: nothing in the suite compared the
compiled derivatives to anything but themselves, and `[1]` on 0.5's new scalar return yields the
first parameter *leaf* rather than an error — a version of this update that kept the index would
have compiled a silently truncated gradient and the only thing standing in its way would have
been an integration test failing somewhere downstream.

The check has now been run by hand twice, once for the update and once for this review. It is in
the suite as of this branch: a new testset in `test/unit/dispatch_variants_unit.jl` compares
`dqdθ`, `dvdθ` and `V_func` against `ForwardDiff` over a flattened parameter vector, at both
bases and both `TEST_TYPES`, and asserts the flattened lengths (`3S`, and `NP` for
`DenseNetBasis`) so the layout `components!` slices out of is pinned as well as the values. The
reference reaches the network through `basis.NN`, so it shares no code with the symbolic path
beyond the network itself and — unlike `NI.∂NN_ansatz_∂params` — does not need the hand-written
ansatz to agree. The Open Issues entry moved to *Added*.

### Not changed

`Project.toml:26`'s `AbstractNeuralNetworks = "0.6.4"`. This package uses neither
`input_dimension` nor `output_dimension`, so the floor is redundant with what
`SymbolicNeuralNetworks` 0.5 already enforces — but it is harmless and it documents the coupling
the CHANGELOG explains, so I left it.

PR #79, opened 2026-08-15
Take GeometricOptimizers from the registry, and read every solve's outcome

michakraus:
Requires **SimpleSolvers 0.12.1** and **GeometricIntegratorsBase 0.6.3**, both registered. No version bump — this package bumps at release time and the entries go under `[Unreleased]`; see *One thing to decide* below.

## The `[sources]` pin is gone

GeometricOptimizers 0.2.0 is in General now, so the git pin goes, and **two of the three entries under *Known issues* close with it**:

- This package can be **tagged** again. A git `[sources]` entry blocks registration in General, which is why it could not be.
- **Julia 1.10 can install it** again. `[sources]` is a Pkg 1.11 feature that 1.10 ignores, which left `GeometricOptimizers = "0.2"` with no candidate and `Pkg` reporting *"Unsatisfiable requirements … no versions left"*. The three 1.10 CI jobs should go green — **not verified locally**, no 1.10 was run here.

The registered GeometricOptimizers 0.2.0 requires SimpleSolvers 0.12, which is what forces the SimpleSolvers bump. The `[sources]` explanation in `benchmark/Project.toml`, which existed only to say why the pin was *not* repeated there, goes with it.

## Reading the solve outcome

SimpleSolvers 0.12 stopped emitting line-search warnings from inside a `NonlinearSolver`'s iteration, so a rejected line search reports to its **caller** through the returned status and to nobody at all if the caller drops it. All three nonlinear solves — the network integrators' shared `integrate_step!`, `CGVINodal`'s and `VISE`'s — move to `solve_with_status!` and hand the status to `check_solver_status`, which GeometricIntegratorsBase 0.6.3 adds as the one place a step's outcome is acted on. That is the function the commented-out stubs in `cgvi.jl` and `vise.jl` had named for years without being able to call. It is silent by default, so **nothing changes in what a run prints**.

`CGVINodal` and `VISE` also move to the **state-taking** form. Both are `GeometricIntegrator`s carrying a persistent `solverstate` that their three-argument call ignored, allocating a `NonlinearSolverState` on every time step; the network integrators already passed theirs.

0.12's other two breaks do not apply: nothing here constructs a `NonlinearSolverStatus`, and the one third-party `LinesearchMethod` in the dependency graph is GeometricOptimizers' `DecayingStatic`, which implements both `solve` and `solve_with_status`.

## `ShallowNet`'s seeding runs through `solve!`

`TrainingMethod` seeding uses `GeometricOptimizers.solve!` instead of a hand-rolled epoch loop, and reads the `OptimizerResult` it returns. The epoch budget becomes `max_iterations`, since `solve!` runs its own loop against `meets_stopping_criteria`, and `warn_iterations = 0` goes with it — reaching a 50 000-epoch budget is the normal outcome here rather than a diagnosis, and at the default of 1000 the warning would print once per dimension per time step.

**The trained seed can differ.** The loop makes the same `solver_step!` calls in the same order, so a run that uses its whole budget is unchanged; but `solve!` may also stop *early*, on the convergence criteria or on a non-finite iterate, where the old loop always burned all `nepochs`. Since this only produces the initial guess for the Newton solve that follows, an earlier stop costs iterations in that solve rather than accuracy in the result. The debug line now reports the epochs actually spent and whether the optimizer converged, instead of asserting the budget.

**Both `DenseNet` loops stay hand-rolled**, deliberately: one breaks early on the loss, the other re-solves the `L3` layer by least squares *inside* every epoch, and `solve!` has a hook for neither.

## Verification

- Full suite passes — **1957 assertions in 7m42s**, run by the pre-push hook.
- The suite resolves entirely from the registry: GeometricOptimizers 0.2.0, GeometricIntegratorsBase 0.6.3, SimpleSolvers 0.12.1, no path or git overrides anywhere.
- `ShallowNet` + `TrainingMethod` is exercised by `test/unit/dispatch_variants_unit.jl`, so the converted loop is covered rather than merely compiled, and `test/integration/shallownet_accuracy.jl` passes unedited — its thresholds did not need retuning.
- **Zero warnings** in the whole run, which is the check that `warn_iterations = 0` landed.

## Open Issues recorded

- **A GeometricOptimizers optimizer has no `solve_with_status!`.** `solve!` does return an `OptimizerResult` carrying the outcome, which is what the `ShallowNet` loop now reads — but `OptimizerResult`, `OptimizerStatus` and `status` are none of them exported, so reading it means reaching past the exported surface.
- **The two `DenseNet` loops check no optimizer status**, only their own loss thresholds, so a non-finite iterate or a diverging optimizer is invisible to both.
- The existing `GradientMethod` + searching-line-search entry is marked **unverified** against SimpleSolvers 0.12, which rewrote the `LinesearchMethod` contract that entry lives in. Nothing here exercises it, because `Static` is still passed explicitly.

## One thing to decide

No version bump is included. `[Unreleased]` now holds a substantial backlog — including the GeometricOptimizers 0.1 → 0.2 migration, whose own notes say the trained weights are not bit-identical to previous releases. Whether that ships as `0.3.1` or `0.4.0` is a judgement call left to the maintainer; the package is taggable either way now that the git pin is gone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,15 +80,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   copy requires SimpleSolvers 0.12, which is what forces the SimpleSolvers bump here. Two of the
   three entries under *Known issues* below close with it — this package can be tagged again, and
   **Julia 1.10 can install it again**, `[sources]` having been the Pkg 1.11 feature that made the
-  three 1.10 CI jobs unsatisfiable. The `[sources]` explanation in `benchmark/Project.toml`, which
-  existed only to say why the pin was *not* repeated there, goes with it.
+  three 1.10 CI jobs unsatisfiable. The `[sources]` explanations in `benchmark/Project.toml` and
+  `scripts/Project.toml`, which existed only to say why the pin was *not* repeated there, go with
+  it.
 
   SimpleSolvers 0.12 is breaking in one way that reaches this package: a `NonlinearSolver` no
   longer emits line-search warnings from inside its iteration, so a rejected line search reports to
   its caller through the returned status and to nobody at all if the caller drops it. See the next
   entry. Its other two breaks do not apply — nothing here constructs a `NonlinearSolverStatus`, and
   the one third-party `LinesearchMethod` in the dependency graph is GeometricOptimizers'
-  `DecayingStatic`, which implements both `solve` and `solve_with_status` and was unaffected.
+  `DecayingStatic`, which implements `solve_with_status` — the side of the rewritten contract that
+  is now the one to implement — and has `solve` derived from it, so it was unaffected.
 
 - **The nonlinear solves use `solve_with_status!`**, and hand the status to
   `check_solver_status`, which GeometricIntegratorsBase 0.6.3 adds as the single place a step's
@@ -116,6 +118,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Since this only produces the initial guess for the Newton solve that follows, an earlier stop
   costs iterations in that solve rather than accuracy in the result. The debug line now reports the
   epochs actually spent and whether the optimizer converged, instead of asserting the budget.
+
+  In practice it does *not* stop early at these settings, and that is by construction rather than
+  by luck: `allow_f_increases` defaults to `true`, so `Adam` overshooting on one step does not end
+  the solve, and `DecayingStatic` is still taking steps of `η₂ = 5e-5` at the horizon — eleven
+  orders of magnitude above the `x_abstol` of `2eps`, which is the gate that would have to fire.
+  What the assessment costs is one extra objective evaluation per epoch: `solve!` builds an
+  `OptimizerStatus` after every `solver_step!`, which the hand-rolled loop did not.
 
   There is no `solve_with_status!` for a GeometricOptimizers optimizer — see *Open Issues*.
 
@@ -715,8 +724,8 @@ release's `Fixed` when they are dealt with.
 
 ### Environment
 
-They are described in full under [Unreleased] → *Known issues*; they are listed here only so that
-this section is the complete index.
+The first three are described in full under [Unreleased] → *Known issues*; they are listed here
+only so that this section is the complete index.
 
 - ~~**The git `[sources]` pin blocks registration in General.**~~ **Closed**: GeometricOptimizers
   0.2.0 is registered and the pin is gone.
@@ -754,10 +763,11 @@ this section is the complete index.
   nonlinear solvers in 0.11 and a state-taking form in 0.12.1, and this package now uses it
   throughout; the optimizer side has no counterpart. `solve!(x, state, opt)` does return an
   `OptimizerResult` carrying the outcome, which is what the `ShallowNet` seeding loop now reads,
-  so nothing is unreachable — but `OptimizerResult`, `OptimizerStatus` and `status` are none of
-  them exported by GeometricOptimizers, so reading it means reaching past the exported surface
-  (`GeometricOptimizers.status(result)`). Either an exported accessor or a `solve_with_status!`
-  would close it.
+  so nothing is unreachable — but `OptimizerResult` and `OptimizerStatus`, and every accessor the
+  seeding loop reads them through (`status`, `isconverged`, `iteration_number`), are none of them
+  exported by GeometricOptimizers, so all four call sites reach past the exported surface
+  (`GeometricOptimizers.status(result)` and friends). Either exported accessors or a
+  `solve_with_status!` would close it.
 
 - **The two `DenseNet` seeding loops check no optimizer status.** They drive the optimizer by hand
   (`increase_iteration_number!` / `solver_step!` / `update!`) because `solve!` has no hook for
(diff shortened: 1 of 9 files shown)


michakraus (commented):
## Review

I read the diff against the installed sources of the three dependencies it moves onto — GeometricOptimizers 0.2.0, SimpleSolvers 0.12.1, GeometricIntegratorsBase 0.6.3 — rather than against the PR description. **The code is correct**; every finding below is in prose that has gone stale or was never right, plus one file the `[sources]` sweep missed. I have pushed fixes for all seven to this branch.

### What I verified

- `check_solver_status` is exported by GeometricIntegratorsBase 0.6.3 (`GeometricIntegratorsBase.jl:86`) and reaches all three call sites through the blanket `using GeometricIntegratorsBase`; the fallback is `check_solver_status(status, int) = status`, so "silent by default" is right, and so is the framing of it as the one place to override.
- `solverstate(int)` is a field on *every* `GeometricIntegrator`, built once as `SolverState(solver)` in the constructor (`integrator.jl:48`). CGVI and VISE really were carrying a state their three-argument call discarded, so that half of the change is a straight fix.
- **"Nothing changes in what a run prints" holds, and for a better reason than the description gives.** `solve!` and `solve_with_status!` are not two code paths: both are one-line wrappers over `_solve_core!` (`nonlinear_solver.jl:341` and `:534`), and `nonlinear_solver_warnings` is *inside* that shared body. There is no way to lose a warning by switching, which is worth knowing since the description reads as though the equivalence were a policy choice.
- The state-taking `solve_with_status!` really is new in 0.12.1 — 0.12.0 has only the two-and-three-argument forms. The compat floor of `0.12.1` is load-bearing, not cosmetic.
- **The `ShallowNet` conversion does not stop early at these settings, by construction rather than by luck**, which I think is a stronger claim than the description makes and worth having written down. `allow_f_increases` defaults to `true`, so `Adam` overshooting on a step does not end the solve; `f_abstol` is `0`; and `DecayingStatic` is still taking steps of `η₂ = 5e-5` at the horizon, eleven orders of magnitude above the `x_abstol` of `2eps` that `x_converged` would need. `meets_stopping_criteria` fires on `iterations ≥ max_iterations` and nothing else here, so exactly `nepochs` `solver_step!` calls happen, in the same order — identical to the old loop. `solve!` skips `update!` on the final iteration only, which does not touch `x`.
- `max_iterations` / `warn_iterations` are valid `Options` keywords and reach it through `Optimizer`'s single `options_kwargs...` level; `warn_iterations = 0` disables the warning, `warn_iteration_number` gating on `> 0`.

### Findings

**1. `scripts/Project.toml` still carries the deleted `[sources]` comment.** The five-line "The GeometricOptimizers git pin is *not* repeated here…" block was removed from `benchmark/Project.toml`, but the byte-identical copy in `scripts/Project.toml` survived and now explains a pin that does not exist. Deleted, and the CHANGELOG entry now names both files.

**2. `src/NonlinearIntegrators.jl` — the `import GeometricOptimizers` comment is wrong, in the old text and more so in the new.** Two things:

- The pre-existing parenthetical "GeometricOptimizers re-exports SimpleSolvers' `solve!` and `Newton` (the same generics, which it `import`s)" is right about `solve!` and wrong about `Newton`. GeometricOptimizers defines **its own** `struct Newton <: OptimizerMethod` (`optimizer_methods.jl:36`) and exports it; it is a different type from SimpleSolvers' `struct Newton <: NonlinearSolverMethod`. That clash is the actual reason `import` is the right call here — and the comment was arguing from the one name where there is no clash.
- The new sentence, "an unqualified `solve!` here would now unambiguously mean the optimizer's", is not true either way round. `import GeometricOptimizers` introduces no names, so an unqualified `solve!` is an `UndefVarError`; and even under `using`, GeometricOptimizers `import`s `solve!` from SimpleSolvers (`GeometricOptimizers.jl:91`), so there is one generic function and no package for it to "mean".

Rewritten to argue from `Newton`, and to say that `solve!` and `update!` are shared generics rather than competing bindings.

**3. `CHANGELOG.md`: `DecayingStatic` does not "implement both `solve` and `solve_with_status`".** `decaying_static.jl` defines only `solve_with_status`; `solve` is derived from it by SimpleSolvers 0.12. That is precisely the new contract the sentence is arguing is satisfied, so getting it backwards undercuts the point. Corrected.

**4. `CHANGELOG.md`, *Open Issues* → *Environment*: the index intro now over-claims.** "The first three are described in full under [Unreleased] → *Known issues*" became "They are described in full…", but the section still has four bullets and the fourth — the Julia 1.13 CI test phase doubling — has no *Known issues* entry. The original wording was precise and the edit had nothing to gain; restored.

**5. `CHANGELOG.md`, the new *no `solve_with_status!` on the optimizer* entry under-lists the unexported surface.** It names `OptimizerResult`, `OptimizerStatus` and `status`, but `shallownet.jl` also calls `GeometricOptimizers.iteration_number` and `GeometricOptimizers.isconverged`, neither of which is exported either. Four call sites reach past the exported surface, not one. Entry extended.

**6. `CHANGELOG.md`: the pre-rename filenames survive in the *Open Issues* standing list.** This PR updated `NonLinear_DenseNet_GML.jl` → `src/nvi/densenet.jl` in the *Upstream* entry but left the same list saying `NonLinear_DenseNet_GML.jl` twice more and `NonLinear_OneLayer_GML` once. Since that list exists so that reviewing the same code twice does not rediscover the same defects, a path that no longer resolves defeats it. Swept the rest (the released sections are historical record and are left alone).

**7. `src/vise/vise.jl`: the comment reads as if the test still does not exist.** "It went unnoticed because `test/unit/vise_unit.jl` was written but never included" — it *is* included, at `test/runtests.jl:39`. As written it invites someone to helpfully re-add it. Reworded to say the inclusion is what would now catch the same slip.

### Notes, not defects

- **`solve!` costs one extra objective evaluation per epoch** relative to the hand-rolled loop: it builds an `OptimizerStatus` after every `solver_step!` (`optimizer.jl:517`), and `OptimizerStatus` takes `value(problem(opt), x)`. That is the price of the assessment and it is worth paying, but with a 50 000-epoch budget per dimension per time step it is not free, and the standing *Julia 1.12 spends hours in type inference* entry sits on exactly this call graph. I have recorded it in the CHANGELOG entry rather than treating it as a problem — no timings re-measured.
- **The version-bump question is yours.** Nothing in the diff decides between `0.3.1` and `0.4.0`; the GeometricOptimizers 0.1 → 0.2 migration already in `[Unreleased]` says the trained weights are not bit-identical to previous releases, which is the fact the decision turns on. Left untouched.
- `check_solver_status` is a no-op by default and this package overrides it nowhere, so there is nothing to test yet. Worth revisiting if a strict override is ever wanted for the network integrators, where a silently non-converged step is currently indistinguishable from a converged one.

PR #82, opened 2026-08-16
Audit: fix five bugs, de-allocate the Newton hot path, consolidate the tests

michakraus:
Audit of the package for correctness, type stability and allocations, plus the test
consolidation that was asked for alongside it.

**2005 tests pass, against 1957 before.** (The pre-push hook reports 1995 + 2 skipped:
Aqua and JET are gated to stable Julia ≥ 1.12, so they run on the `1.12` CI job and can
be forced locally with `NI_STATIC_ANALYSIS=true`.)

## Bugs

- **`VISE.integrate!` corrupted its output on any restart.** It sized `internal_values`
  and `each_step_solution` as `n₂-n₁+1` but indexed them by `n`, so every `n₁ > 1` left
  the first `n₁-1` slots `#undef` and ran off the end. This is the same bug that had
  already been found and fixed in the network integrators' `integrate!` — comment and
  all; the VISE copy of the loop was never updated.
- **`box_init_plain` reseeded Julia's global RNG as a side effect of a default
  argument.** The keyword defaulted to the *expression* `Random.seed!(1)`, evaluated on
  every call that omits it. DenseNet's LSGD seed calls it three times in a row, so all
  three layers were drawn from one freshly-reseeded stream — correlated, not
  independent — discarding the `Random.seed!(42)` the caller had just done. It also
  defaulted to `Float32` while every call site omitted the type, so a `Float64` DenseNet
  drew its weights at single precision. That contradicts the package's central
  no-silent-upcast invariant and was invisible to a test checking the `eltype` of the
  result.
- **The two reversible integrators printed the full residual vector by default.** The
  `println` sits inside `residual!`, gated on `show_status`, which defaulted to `true` —
  and `residual!` runs once per Newton iteration *and* once per ForwardDiff Jacobian
  column, with `b` a vector of `Dual`s. The non-reversible siblings had already had
  those lines removed.
- **`VISE` and `CGVINodal` trait functions never reached the framework.**
  `isexplicit`/`isimplicit`/`issymmetric`/`issymplectic` were defined unqualified. None
  of the four is imported into this module, so each definition created a *new*,
  shadowing `NonlinearIntegrators.isexplicit` rather than extending
  `GeometricIntegratorsBase`'s generic, and the framework kept answering `missing`. It
  matters most for `CGVINodal`, where `issymplectic = true` is a real claim about the
  continuous-Galerkin construction and was the one property downstream code could have
  selected the integrator on.
- **`VISEBasis` built its callables with `eval`.** `Symbolics.eval(build_function(…))`
  adds methods in a newer world age than the caller's, so constructing a basis and
  evaluating it inside one function body raised *"the applicable method may be too
  new"*. It only appeared to work because the single test built the basis at top level,
  where world age advances between statements.

## Type stability

`CacheType` built its return type from **runtime** fields (`method.basis.S`,
`method.common.extrapolation_substep`), so it returned
`ShallowNetCache{Float64, _A, 8, _B} where {_A, _B}` — not concrete. That is the type the
framework's `CacheDict` lookup asserts against, so `cache(int, ST)` came back abstract and
every value read out of it was `Any`. Those parameters were phantom: no field type
mentioned them and nothing dispatched on them. They are ordinary constructor arguments
now.

With that, and `ForwardDiff.gradient!` in place of `gradient` (whose chunk size — and
hence return type — is chosen from a runtime length), JET goes from 13 runtime dispatches
in `ShallowNetAutodiff`'s `residual!` to **none, for all four integrators**.

Also fixed: `flatten_params` (`Vector{Any}` plus a splatted `vcat`, in DenseNet's hot
loop), the boxed `sw` capture in `oga_fit`, and VISE's untyped fields and `mat = []`
builders.

## Allocations

Bytes per `residual!` call (Float64, S = 4, R = 8, D = 1):

| integrator | before | after | |
|---|---|---|---|
| `ShallowNet` | 21 344 | 11 424 | 1.9× |
| `ShallowNetReversible` | 26 176 | 11 424 | 2.3× |
| `ShallowNetAutodiff` | 211 968 | 51 584 | 4.1× |
| `ShallowNetAutodiffReversible` | 216 800 | 51 584 | 4.2× |

Most of it was work that was thrown away: six cache arrays written on every residual and
**never read anywhere**, along with the two `ForwardDiff.gradient` calls that existed only
to fill them; `DVDθ` evaluated *twice* per quadrature node although one call returns both
halves; `apply_NN` materialising three slices and three broadcast temporaries per call,
inside a gradient and again inside a nested gradient-of-a-derivative; `ps_vec` allocated
up to `2D+1` times per call across three loops recomputing the same gather; and `s̃`,
allocated in all seven caches and never read or written.

## Tests

The five per-integrator unit files were structurally identical — the same
three-element extrapolation list declared five times, the same accuracy block and
cross-product loop varying only a constructor and one tolerance. Two of them differed in
27 of 44 lines, every one a type-name substitution plus a single number. They are now one
table-driven file, with the shared fixtures in `testsetup.jl`.

Bases are memoised: the suite performed ~72 SymbolicNeuralNetworks code-generation runs
to obtain about two distinct objects per element type, and the pre-push hook pays that on
every push.

New gates: per-integrator `@allocated` budgets and `@inferred` on `residual!`; Aqua's
full suite; and a **D = 2 layout guard** for the network integrators — every layout
mistake between `components!`, `residual!` and `update!` collapses to the identity at
D = 1, which is exactly why `cgvi_unit.jl` already carried the equivalent guard. VISE's
test went from one Float64 step asserting only finiteness to a five-step accuracy
assertion (its ansatz is exact for the harmonic oscillator), a restart guard, and trait
checks.

## Two things to look at

**JET is a developer script, not a suite assertion.** `test/quality/jet_residual.jl`
reports clean in every environment I could construct — plain session, `Pkg.test`-equivalent
environment, with and without `--check-bounds=yes`, `--pkgimages=no`, `-O0`, `-g1`. Inside
the `Pkg.test` process itself it reports twelve dispatches whose shape (`view(r, :, d)`
inferring as `SubArray{…} where {N,I}`) looks like inference widening under accumulated
specialisations. I could not reproduce or explain that, so I did not ship a gate that
fails for reasons I can't account for. The reasoning is in `test/quality/aqua_jet.jl`.

**Aqua ambiguities: 8 → 3.** Five were ours and are fixed by giving the package-local
DT-form update its own name (`update_solution!`) instead of making it a fourth `update!`
method. The remaining three come from following the framework's own documented extension
signature and collide with `GeometricOptimizers.update!(::BFGSState, …)`; the ambiguous
call is not reachable. They are excluded by *function*, so a new ambiguity anywhere else
still fails.

## Not done

- Merging the `components!` / `residual!` bodies of the shallow-net pairs. The caches
  are merged (they were byte-identical bar one line) and the D = 2 guard is in place as
  scaffolding, but the bodies themselves are untouched — the riskiest item, and better
  done with room to iterate.
- Hoisting the OGA seed's `Ψ` (8.8 MB at the default `dict_amount = 50000`) out of the
  per-component and per-step loops. It is the single largest allocation in the package
  and is unchanged; `Ψ` does not depend on the solution component it is rebuilt for.


michakraus (commented):
Read the whole diff against `main`. **The substance holds up.** Spot-checking the load-bearing claims:

- The `update!` → `update_solution!` rename is safe: `GeometricIntegratorsBase`'s documented extension point is the 4-arg `(sol, params, x, int)` form (`euler/implicit_euler.jl:96`), and nothing in the framework calls the DT-form. So the five ambiguities really were self-inflicted and really do go away.
- The merged `components!` loops in the symbolic pair are aliasing-safe: every `DQDθ`/`DVDθ` result is copied out before the next call, so a compiled kernel returning a reused buffer cannot bite.
- The `q̃plain` hoist is behaviour-preserving — the old code read `cache(int).q̃[d]` at the same point, and `q` enters the ansatz only through `t·q`, so the parameter gradients genuinely do not see it.
- The `CacheType` phantom-parameter removal is correct, and `hasfield(typeof(cache(int)), :stage_values)` does now fold.
- `VNN_ansatz`'s closed form keeps each `ForwardDiff.derivative` strictly nested inside one tag level, so it does not reintroduce the ordering error the docstring describes.

No wrong-answer bug found. What follows is one behavioural slip, one dead knob whose CHANGELOG entry is not true, a documented command that cannot run, and some dead code and stale comments that an audit PR should not be leaving behind. All are fixed in the commits that follow this review.

---

## A. The merge conflict, and what merging `main` invalidates

**A1 — `CHANGELOG.md`.** Both branches appended to `## [Unreleased]`. Resolved as a union: `main`'s OGA comparison/adapter-layer note under *Changed*, its Precision-page compat correction under *Fixed*, the docs-deploy entry (from the shared merge base) exactly once.

**A2 — `docs/src/oga/usage.md:139`.** This one only becomes wrong *because* of the merge, which is why it is worth calling out separately. #81 added:

> Per-step diagnostics are printed by default: `show_status = true` … Set `show_status = false` to silence it.

This PR flips that default to `false`. The paragraph is now inverted, and `main`'s CHANGELOG entry drops its "(on by default)" aside.

## B. Correctness

**B1 — `src/vise/vise.jl:372` rebinds a cache slot instead of filling it.** In `record_finer_solution!` the `[:]` came off the **left** side rather than the right:

```julia
tem_W[d] = x[start_idx:start_idx+W_size-1]      # replaces the cache's buffer
```

Everywhere else in this PR the same line became `copyto!(tem_W[d], view(x, …))` — including `components!` twelve lines up, which is what makes this read as a slip rather than a decision. As written it allocates a fresh vector per DOF per step and silently swaps out the cache's own array.

**B2 — VISE's `record_grid_points` is dead, and the CHANGELOG claims it is not.** `VISECache{ST,R}(W_sizes, ics; record_grid_points = 41)` is never passed a value: `VISE` has no such field, and `Cache{ST}(problem, ::VISE)` forwards only `kwargs...`, which the framework calls with none. `record_finer_solution!` still hardcodes `reshape(collect(0:1/40:1), 1, 41)`. So the entry

> The `41`-row `stage_values` buffer now honours `record_grid_points` like the network integrators.

is not true today. Two further consequences: that grid is `Float64` whatever `ST` is, where every network integrator builds `range(zero(T), one(T), N_plot)`; and if anyone ever *did* pass a value ≠ 41, the buffer and the loop would disagree — a `BoundsError` waiting behind a keyword that currently cannot be reached.

**B3 — `test/quality/jet_residual.jl` cannot be run the way its own header says.** Two separate errors:

- *"Driven as a subprocess by `aqua_jet.jl`"* — it is not; `aqua_jet.jl` only `@test_skip`s and points the reader here.
- `julia --project=test test/quality/jet_residual.jl` begins with `using NonlinearIntegrators`, which `test/Project.toml` **deliberately does not list** — its own header comment explains why. The command fails with *"Package NonlinearIntegrators not found"*.

Given that this file is the PR's answer to "why is JET not a gate", the one instruction for reproducing its result needs to work.

## C. Dead code the audit missed

**C1 — `AutodiffShallowNetCache.X` is a second `s̃`.** Allocated in the cache, bound as `local X = cache(int, ST).X` in all four autodiff `components!`/`residual!` bodies, and indexed nowhere — `grep -n 'X\['` over both autodiff files is empty. (`SymbolicShallowNetCache.X` *is* read, by the two symbolic integrators; that one stays.) Exactly the class of thing the `s̃` removal was about.

**C2 — dead locals** in both autodiff `components!`: `local p = cache(int, ST).p̃` and `local NN = method(int).basis.NN`. These integrators never touch the `Chain`, and `p̃` is computed in `update_solution!`.

## D. Comments and docstrings that say the wrong thing

- **D1 — `src/vise/vise.jl:233`:** ``# `tem_W[d]`, not `tem_W[d]``` — the second should be `tem_W[d][:]`.
- **D2 — `src/nvi/utilities.jl:32`:** the `flatten_params!(dest, params) -> dest` docstring is attached to `@generated function _param_arrays`, and its `@ref` to `optimizer_params` reads as if `_param_arrays` were the documented function. `flatten_params!` itself has none.
- **D3 — `src/nvi/network_integrator_core.jl:420`** says "Four of the eight ambiguities"; the CHANGELOG, `aqua_jet.jl` and the PR description all say five. Five is right (8 → 3).
- **D4 — dangling references to `ShallowNetCache`**, a type this PR deletes: `src/cgvi/cgvi.jl:104` and `src/nvi/densenet.jl:58` both say "see the note on `ShallowNetCache`".
- **D5 — `test/quality/inference_and_allocations.jl`** claims `@inferred` on `residual!` catches internal instability. It does not: `@inferred` compares the inferred return type against `typeof(result)`, and `residual!` returns `nothing`, which infers as `Nothing` even with runtime dispatch throughout the body. The byte budget beside it is the assertion that actually bites — worth saying so, since the PR presents the two as a pair.

## E. Test-suite structure

**E1 — `test/quality/inference_and_allocations.jl` reads `NETWORK_INTEGRATORS`**, a `const` defined at the top of `test/unit/network_integrators_unit.jl`. `quality/` therefore only works if `unit/` ran first. This PR moved `EXTRAPOLATIONS`, `ho_accuracy_problem`, `oga_testcase`, `accuracy_guard` and `dispatch_case` into `testsetup.jl` for precisely this reason; the integrator table is the largest shared fixture and is the one that stayed behind.

**E2 — stale committed manifests:** `docs/Manifest.toml:978` and `benchmark/Manifest.toml:1587` still list `Zygote` among NonlinearIntegrators' dependencies.

---

## On the two items flagged "look at this"

**JET as a script rather than a gate — agreed.** A check that fails only inside `Pkg.test`, for a reason that reproduces nowhere else, is worse than no check. The `@allocated` budgets are the right substitute because they measure the consequence rather than the mechanism. The only thing to fix is B3, so the escape hatch actually opens.

**The Aqua exclusion — agreed, and correctly scoped.** Excluding by function rather than disabling `test_ambiguities` is the distinction that matters, and `Aqua.test_all(…; ambiguities = false)` followed by a separate `test_ambiguities(…; exclude = …)` is the right way to express it.

## On "Not done"

Both deferrals are the right call. Merging the `components!`/`residual!` bodies of the shallow-net pairs without the D = 2 guard in place would have been the risky ordering; with the guard landed first, that work has a net under it. Same for hoisting the OGA seed's `Ψ` — 8.8 MB is the biggest single allocation in the package, but it is not on the Newton path this PR is about.

src/vise/vise.jl
```diff
@@ -339,7 +369,7 @@ function record_finer_solution!(sol, int::GeometricIntegrator{<:VISE})
 
     start_idx = 1
     for (d, W_size) in enumerate(W_sizes)
-        tem_W[d][:] = x[start_idx:start_idx+W_size-1]
+        tem_W[d] = x[start_idx:start_idx+W_size-1]
```
michakraus: B1: the `[:]` came off the **left** side here, not the right. This rebinds the cache slot to a freshly allocated copy of the slice instead of filling the buffer that is already there — one allocation per degree of freedom per step, and the cache quietly ends up holding different arrays than the ones it was constructed with.

The sibling line in `components!` (line 216 of this file) got the intended treatment:

```julia
copyto!(tem_W[d], view(x, start_idx:start_idx+W_size-1))
```

src/vise/vise.jl
```diff
@@ -102,9 +116,9 @@ struct VISECache{ST,R} <: IODEIntegratorCache{ST}
 
         tem_W = create_boundary_derivative_vector(ST, D, W_sizes)
 
-        stage_values = zeros(ST, 41, D)
+        stage_values = zeros(ST, record_grid_points, D)
```
michakraus, line 159: B2: `record_grid_points` never arrives here. `VISE` has no such field, and `Cache{ST}(problem, ::VISE)` (line 127) forwards only `kwargs...`, which the framework calls with none — so `stage_values` is always 41 rows, and `record_finer_solution!` still hardcodes `reshape(collect(0:1/40:1), 1, 41)` to match.

That makes the CHANGELOG line — *"The `41`-row `stage_values` buffer now honours `record_grid_points` like the network integrators"* — untrue, and it leaves a `BoundsError` behind a keyword nobody can currently reach: pass 21 and the buffer and the loop disagree.

The hardcoded grid is also `Float64` regardless of `ST`, where the network integrators build `range(zero(T), one(T), N_plot)`.

test/quality/jet_residual.jl
```diff
@@ -0,0 +1,70 @@
...
+#     specialisations of the same functions, and Julia's inference widens under that load —
+#     `view(r₀, :, d)` comes back as `SubArray{Float64,N,Matrix{Float64},I,true} where {N,I}`
+#     rather than the concrete type it infers in a clean session. The reports that follow are
+#     artefacts of the analysis context, not of the shipped code.
+#   * `JET.test_package` is not usable either: it re-analyses the package from source in the
+#     *active* environment, which under `test/Project.toml` does not carry the package's own
+#     dependencies, and reports 108 "Package GeometricEquations not found" toplevel errors.
+#
+# A fresh process analysing `residual!` at the concrete argument types is what actually answers
+# the question "does the hot path dispatch dynamically". Exits 0 if clean, 1 otherwise, printing
```
michakraus, line 16: B3: neither half of the header holds.

*"Driven as a subprocess by `aqua_jet.jl`"* — it is not; `aqua_jet.jl` `@test_skip`s and points the reader back here.

And `julia --project=test test/quality/jet_residual.jl` cannot work: line 18 is `using NonlinearIntegrators`, which `test/Project.toml` deliberately omits (its own header explains why). It fails with *"Package NonlinearIntegrators not found"*.

This file is the PR's answer to "why is JET not asserted", so the one command for reproducing its result needs to run.

src/nvi/utilities.jl
```diff
@@ -29,34 +29,55 @@ function simpson_quadrature(N::Int, ::Type{T}=Float64) where {T}
...
-    
-    nodes = 0.5 .* ([-1 .* points...,reverse(points)...] .+ 1)
-    weights = 0.5 .* [weight...,reverse(weight)...]
-    return (nodes = nodes, weights = weights)
-end
+"""
```
michakraus, line 32: D2: this docstring documents `flatten_params!` but is attached to `_param_arrays`, so `?flatten_params!` finds nothing and `?_param_arrays` returns a signature it does not have. The `@ref` to `optimizer_params` compounds it — read against `_param_arrays` it says the `@generated` reason applies to a function whose docstring is about copying into `dest`.

src/nvi/network_integrator_core.jl
```diff
@@ -208,21 +412,33 @@ function GeometricIntegratorsBase.residual!(
...
-# Default DT-form update!: copy q̃/p̃ from cache into solution.
-# ShallowNetAutodiff and ShallowNetAutodiffReversible override this with physics-specific momentum.
-function GeometricIntegratorsBase.update!(
-        sol, params, int::GeometricIntegrator{<:NetworkIntegratorMethod}, DT)
+# `update_solution!` is this package's own function, not an extension of the framework's
+# `update!`. Splitting the update into "run components!, then write the cache into `sol`" is a
+# local convention — `GeometricIntegratorsBase` has no such second form — and defining it as a
+# fourth `update!` method meant a signature of the shape `(Any, Any, GeometricIntegrator, Any)`,
+# which is ambiguous against the framework's own `(Any, Any, Any, GeometricIntegrator)`
+# (`explicit_euler.jl`). Four of the eight ambiguities Aqua reported were exactly that pair, for
```
michakraus: D3: "Four of the eight" — the CHANGELOG, `test/quality/aqua_jet.jl` and the PR description all say five, and five is what reconciles with 8 → 3.

michakraus:
Two corrections and one addition to the review above, found while applying the fixes.

**Added finding — the Documenter build fails on this branch.** I missed this on the first pass because it is not visible in the source diff; it only shows up when `makedocs` runs. Two of the docstrings this PR adds link to bindings that have none, and `index.md` pulls everything in through `@autodocs`:

```
Error: Cannot resolve @ref for md"[`VISE`](@ref)" in docs/src/index.md.
│ - No docstring found in doc for binding `NonlinearIntegrators.VISE`.
Error: Cannot resolve @ref for md"[`NN_ansatz`](@ref)" in docs/src/index.md.
│ - No docstring found in doc for binding `NonlinearIntegrators.NN_ansatz`.
ERROR: LoadError: `makedocs` encountered an error [:cross_references] -- terminating build before rendering.
```

`git log -S` puts both refs in `b41bbea`, so the Documenter CI job would have gone red on this PR. Fixed by giving both targets a docstring rather than by dropping the links — `VISE` is an exported public type that had none at all, and its three-element `integrate` return was not stated on any page.

**Correction to E2.** I called `docs/Manifest.toml` and `benchmark/Manifest.toml` "stale committed manifests". They are not committed — `.gitignore:35` excludes `Manifest.toml` repo-wide, and `git ls-files` returns nothing for either. The Zygote entry I was reading is in the local working copies, which regenerate on `Pkg.instantiate()`. Nothing to fix; disregard that item.

**Correction to D3.** I asserted five without showing the arithmetic. For the record it is one ambiguity per DT-form `update!` definition — the default in `network_integrator_core.jl` plus the four overrides in `ShallowNetAutodiff`, `ShallowNetAutodiffReversible`, `CGVINodal` and `VISE`. That is the count now in the comment.

---

Everything else is applied as described. Verification on this machine (Julia 1.13.0-rc2):

- `Pkg.test`: **2000 pass, 2 skip, 0 fail**. The 1995 this branch had plus the 5 new `record_grid_points` assertions — so moving `NETWORK_INTEGRATORS` into `testsetup.jl` dropped nothing.
- `NI_STATIC_ANALYSIS=true Pkg.test`: **2010 pass, 1 skip**. Aqua runs and passes, ambiguity exclusion included; only the JET `@test_skip` remains. (Both gates skip by default here because 1.13.0-rc2 is a prerelease, which is the documented behaviour.)
- `jet_residual.jl` under the command now in its own header: clean for all four integrators.
- `include("docs/make.jl")`: exit 0, no cross-reference errors.

PR #83, opened 2026-08-16
Remove CGVINodal: it moves to GeometricIntegrators

michakraus:
`CGVINodal` is a plain **linear** continuous Galerkin variational integrator. This package is for the *nonlinear* ones — the integrators whose ansatz is a neural network or a symbolic expression — and `CGVINodal` was only ever here as the linear reference those are compared against. It belongs upstream next to `CGVI`, and it is now there:

* **JuliaGNI/GeometricIntegrators.jl#219** — merged, and released in **GeometricIntegrators v0.18.3** (registered in General). It also removes the duplication between the two CGVI variants, putting everything they share behind one `CGVIMethod` supertype in `integrators_cgvi_common.jl`.

## What moves with it

Everything `test/unit/cgvi_unit.jl` asserted:

| test | now |
|---|---|
| per-precision accuracy (`Float64`, `Float32`), no silent upcast | upstream, for **both** `CGVI` and `CGVINodal` |
| `D = 2` `CoupledHarmonicOscillator` layout regression (`q` and `p`) | upstream, for both |
| trait sweep (`isexplicit`/`isimplicit`/`issymmetric`/`issymplectic`/`isiodemethod`) | upstream, for both |
| construction smoke test over `TEST_TYPES` | superseded by the accuracy runs upstream |

The coverage is wider upstream than it was here, not narrower — every property that was asserted for `CGVINodal` alone is now asserted for `CGVI` too. Upstream additionally rejects a basis whose nodes do not include the interval endpoints, which was previously only documented as a requirement.

## What stays

The `D = 2` guard in `test/unit/network_integrators_unit.jl` is the network integrators' own, so `GeometricProblems.CoupledHarmonicOscillator` is still imported in `test/testsetup.jl`. Comments that cited `CGVINodal` as an example — the `Newton`-clash note in `src/NonlinearIntegrators.jl`, the `update_solution!` note in `src/nvi/network_integrator_core.jl`, the argument-order note in `src/vise/vise.jl` — are repointed rather than deleted.

## Migration

```julia
using GeometricIntegrators: CGVINodal   # was: using NonlinearIntegrators: CGVINodal
```

Requires GeometricIntegrators v0.18.3 or later. The constructor, the coefficients and the numerics are unchanged, so results are identical.

**No dependency on GeometricIntegrators is added.** Nothing left in this package refers to the integrator, so this merges independently of the upstream PR and the test suite stays green either way.

## Review follow-ups

Two commits on top of the removal, from the review:

* `cf21e0a` — the CHANGELOG asserted the new home in the present tense and told readers to write `using GeometricIntegrators: CGVINodal` while #219 was still open, i.e. while that line threw `UndefVarError`; it was reworded to state the condition. Also: `test/testsetup.jl`'s `lobatto` helper was left dead by this PR (its only two call sites were the deleted `cgvi_unit.jl` and the deleted `CGVINodal` smoke block) and is removed, plus two comment fixes.
* `4026457` — #219 has since merged and shipped in v0.18.3, so the conditional wording is replaced by the concrete version requirement above.

Full suite green: 1982 pass, 2 broken (pre-existing).


michakraus (commented):
The deletion itself is complete and I could not find a loose end in the code. What I checked:

* **No dangling references.** `grep -rniE "cgvinodal"` over `src/`, `test/`, `docs/`, `benchmark/` and `scripts/` comes back empty. The `CGVI(...)` calls in `scripts/vise_plot.jl` and `scripts/test_vise.jl` are GeometricIntegrators' own `CGVI`, and `obsolete/BSpline/CGVI_SplineBasis.jl` is the retired variant — neither is affected.
* **No dependency goes stale.** `SMatrix` was the only StaticArrays name unique to `cgvi.jl`; `SVector` is still used by the four shallow-net methods and `VISE`, so `using StaticArrays` stays honest, and `QuadratureRules`/`CompactBasisFunctions` are load-bearing for the network and VISE bases. Aqua's `stale_deps` would have caught it otherwise, and it is green.
* **The repointed comments are accurate.** `default_solver(::VISE) = Newton()` really does exist (`src/vise/vise.jl:83`), so the `Newton`-clash note in `src/NonlinearIntegrators.jl` still has a live referent; `CoupledHarmonicOscillator` really is still used (`test/unit/network_integrators_unit.jl:115`), so the `import` in `testsetup.jl` and its rewritten justification hold. The "five ambiguities" count in `network_integrator_core.jl` still adds up — one default plus three current overrides plus `CGVINodal`, now marked historical.
* **All 13 CI jobs pass**, including nightly and `^1.13.0-0`, plus the docs build.

Four things to fix. One of them matters.

---

### 1. The migration instruction does not work yet

[JuliaGNI/GeometricIntegrators.jl#219](https://github.com/JuliaGNI/GeometricIntegrators.jl/pull/219) is **open**, not merged. But the CHANGELOG asserts availability in the present tense —

> **`CGVINodal` is removed. It now lives in GeometricIntegrators** […]
> Replace `using NonlinearIntegrators: CGVINodal` with `using GeometricIntegrators: CGVINodal`.

— and that second line throws `UndefVarError` today, and will keep throwing until #219 both lands *and* ships in a tagged GeometricIntegrators release. The module tombstone at `src/NonlinearIntegrators.jl:117-119` has the same problem ("Reach it as `GeometricIntegrators.CGVINodal`").

The PR body is careful about this for the *merge* order ("this merges independently of the upstream PR"), which is right — nothing here needs #219. But it is not true for the *release* order: `CGVINodal` was exported from the registered v0.2.0, so the next tag is what takes it away from users, and at that moment the replacement has to exist. Reword to state the condition rather than assert the fact, and record the constraint: do not tag the next NonlinearIntegrators release before a GeometricIntegrators release containing #219 exists. I will also add the `#219` link to the two comments that mention the move without one (`src/nvi/network_integrator_core.jl:424`, `test/unit/network_integrators_unit.jl:89`), so the claim is checkable from the source.

### 2. `test/testsetup.jl:57` — `lobatto` is now dead

```julia
lobatto(::Type{T}, R = 4) where {T} = QuadratureRules.LobattoLegendreQuadrature(T, R)
```

Its only two call sites were `test/unit/cgvi_unit.jl` and the `CGVINodal` block in `test/smoke/methods_smoke.jl`, both deleted here; `grep -rn "lobatto" test/` now matches the definition and nothing else. The sibling `gauss` stays — it is used throughout. Removing it, since the PR is otherwise thorough about not leaving CGVI-only scaffolding behind.

### 3. `test/unit/network_integrators_unit.jl:87-90` — the replacement comment is circular

> The guard exists here because that class of bug was first found in the linear reference integrator this package used to carry; it moved to GeometricIntegrators together with `CGVINodal`, whose `test/integrators/galerkin_integrators_tests.jl` now runs it for both CGVI variants.

"the linear reference integrator this package used to carry" *is* `CGVINodal`, so "together with `CGVINodal`" reads as though there were two things, and "whose" then dangles onto a name that has not been introduced as the antecedent. Naming it once fixes both.

### 4. `src/vise/vise.jl:385` — reflow

Dropping "`cgvi/cgvi.jl` and" left the line breaking mid-phrase ("as in the network integrators' shared / `integrate_step!`"). Cosmetic; reflowing while I am in the file.

---

### Not a finding, but worth stating

Between this merge and #219 landing, `CGVINodal` is tested nowhere: the accuracy runs, the `D = 2` layout regression and the trait sweep all leave here in this PR and only arrive upstream when #219 does. That is an accepted consequence of splitting the two PRs rather than a defect — the code is in this repo's history, and #219 already carries the wider version of all three (both `CGVI` and `CGVINodal`, plus a rejection of a basis whose nodes miss the interval endpoints, which was only documented before). It is the same window the release-ordering note above is about.

PR #86, opened 2026-08-20
Take the parameter container from NeuralNetworkParameters, and SymbolicNeuralNetworks 0.6

michakraus:
[`AbstractNeuralNetworks` 0.7](https://github.com/JuliaGNI/AbstractNeuralNetworks.jl/pull/33) moved
the parameter container out to
[`NeuralNetworkParameters`](https://github.com/JuliaGNI/NeuralNetworkParameters.jl), where the type
is called `NetworkParameters`, and removed the old name outright rather than leaving an alias — so
that one type has one name across the ecosystem. An alias was tried upstream and dropped: it could
not be exported without colliding with the module name, and unexported it did not preserve source
compatibility for `using AbstractNeuralNetworks` either. Code written against 0.6 has to be edited
whichever way.

This package follows. `NetworkParameters` is imported from the package that defines it, and the name
is used at every call site — three method signatures and the wrapper construction in
`src/nvi/utilities.jl`, the per-dimension parameter set in `shallownet.jl`,
`shallownet_reversible.jl` and `densenet.jl`, and the `L3` merge in `DenseNet`'s LSGD loss. Nothing
in this package's own exported surface changes, and the numerics are untouched.

The import is selective: `NeuralNetworkParameters` also exports `flatten`/`unflatten`, and the
flat-vector conversions in `nvi/utilities.jl` — `optimizer_params` and `network_params` — are this
package's own, over a layout one level shallower than a network's. The module is named alongside the
type because the `params` accessor is taken qualified off it; `AbstractNeuralNetworks` re-exports
`params`, so both spellings are one binding, but the type and its accessor now read as coming from
the same place.

`SymbolicNeuralNetworks` moves to **0.6** in the same change. This is one constraint with the
`AbstractNeuralNetworks` bump, not two: SNN 0.6 is the release that tracks ANN 0.7, and SNN 0.5 pins
ANN `0.6.4 - 0.6`, so the pair moves together or the environment does not resolve. Nothing else SNN
0.6 changes reaches here — it renames nothing in its exported surface, and the one name it drops,
`QPTOAT`, was never used in this package.

`NeuralNetworkParameters` enters the tree as a dependency of both `AbstractNeuralNetworks` 0.7 and
`SymbolicNeuralNetworks` 0.6. It is listed directly, with a `[compat]` entry, because this package
names the type itself rather than only passing parameter sets through; the test environment lists it
too, since the unit tests construct parameter sets by hand. (The second commit's message says it was
"already a direct dependency" — it was not a dependency at all before the bump.) `benchmark/` takes it *in place of* `AbstractNeuralNetworks`:
`compare_derivative_backends.jl` was the only file under `benchmark/` using that package at all, and
only for the parameter container.

Version goes to **0.4.0** — the dependency floor moving is breaking for anything holding
`AbstractNeuralNetworks` 0.6 alongside this package.

## Verification

Against the registry, nothing dev'd:

```
julia --project -e 'using Pkg; Pkg.update(); Pkg.test()'
```

resolves `AbstractNeuralNetworks v0.7.0`, `NeuralNetworkParameters v0.1.1`,
`SymbolicNeuralNetworks v0.6.0`, and gives **1982 pass, 2 broken** — the two `@test_broken` markers
already on `main` — in 10m03s, including the Aqua and JET quality testsets.

`benchmark/` instantiates against the same three versions, and `kernel_params` — the one function
under `benchmark/` that names the container — returns a `NetworkParameters` whose layers match the
flat vector it is built alongside. The full `quick` sweep and `docs/make.jl` were not run locally;
the docs job failed on resolution alone, and both environments now resolve, so CI covers them.

## CI: the package servers, not the bounds

Every job on this branch died in `Pkg.resolve` before running a test, and the compat bounds were
not the reason. The package servers are serving a registry that predates the releases those bounds
name — as of 2026-08-24 the one behind `https://pkg.julialang.org/registries` stops at
`AbstractNeuralNetworks` 0.6.4, `SymbolicNeuralNetworks` 0.5.0 and `NeuralNetworkParameters` 0.1.0,
against 0.7.0 / 0.6.0 / 0.1.1 in General, the first of which was registered more than a day
earlier.

`JULIA_PKG_SERVER: ""` in `CI.yml` and `Documenter.yml` makes Pkg clone General over git and fetch
packages from their own repositories, which resolves the same environment that resolves locally.
`SymbolicNeuralNetworks` hit this first and carries the same block. It costs speed — no CDN, no
registry tarball — and comes out once the servers catch up. A server's `/meta` reports a
`last_registry_update` that is a heartbeat and not the content date (the one serving 0.6.4 claimed
16:50 on 2026-08-23, hours after 0.7.0 was registered), so the comment in both workflows records
how to check the served registry itself.


michakraus (commented):
Reviewed and updated for the final shape of AbstractNeuralNetworks#33.

## What the CI failure was

Every job, identically: `Unsatisfiable requirements detected for AbstractNeuralNetworks — restricted to versions 0.7 … no versions left`. The compat floor against a registry that only has 0.6.4. It is a gate, not a defect: nothing in this diff can turn it green until `AbstractNeuralNetworks` 0.7.0 is registered. Verified locally instead — `Pkg.test` against a dev'd `AbstractNeuralNetworks` at the #33 branch: **1982 pass, 2 broken, 0 fail**, matching main.

## What changed since

`AbstractNeuralNetworks` 0.7 dropped the `NeuralNetworkParameters` alias entirely rather than leaving it unexported. It never preserved source compatibility for `using AbstractNeuralNetworks` — being unexported, that still raised `UndefVarError` — and keeping it would have left one type with two names across four packages. So the name is `NetworkParameters`, imported from the package that defines it:

```julia
import NeuralNetworkParameters: NetworkParameters
```

The comment explaining the import is rewritten accordingly; it described the alias as "deliberately unexported so that every user of it says where it came from", which is no longer the arrangement.

The import stays selective for a reason worth writing down: `NeuralNetworkParameters` exports `flatten`/`unflatten`, and the flat-vector conversions in `src/nvi/utilities.jl` — `optimizer_params` and `network_params` — are this package's own, over a layout one level shallower than a network's.

Renamed at all 20 call sites in `src/`.

## Three things this PR's diff had missed

- `test/unit/dispatch_variants_unit.jl:87` and `:126` construct parameter sets through the *qualified* `AbstractNeuralNetworks.NeuralNetworkParameters(…)`. A change to an import line does not reach a qualified reference, so those two would have failed under 0.7 whichever way the alias went.
- `test/unit/optimizer_params_unit.jl:9` imports the name from `AbstractNeuralNetworks` on its own line.
- `NeuralNetworkParameters` was already in `[deps]` but had no `[compat]` entry, and the test environment (`test/Project.toml`) did not list it although the unit tests construct parameter sets by hand. Both added.

The comment in `test/testsetup.jl:12` also still claimed `NeuralNetworkParameters` was one of the three names taken from `AbstractNeuralNetworks`; corrected.

## Merge order

Blocked on AbstractNeuralNetworks#33 being merged and 0.7.0 registered. Nothing else.

PR #92, opened 2026-08-25
Move to GeometricOptimizers 0.5 and SymbolicNeuralNetworks 0.7

michakraus:
Three compat bounds, which have to move together.

```
GeometricOptimizers    = "0.2, 0.3, 0.4"  →  "0.5"
NeuralNetworkParameters = "0.1"           →  "0.2"
SymbolicNeuralNetworks  = "0.6"           →  "0.7"
```

`GeometricOptimizers` 0.5.0 drops the `ParameterHandling` shim and lets `NeuralNetworkParameters` do the flattening, so the 0.2 container is a requirement rather than a choice — 0.4 and earlier want the 0.1 container. And `SymbolicNeuralNetworks` 0.7.0 is the first release that permits the 0.2 container, where 0.6.0 allows only 0.1. Any one of the three alone is unsatisfiable.

## No code changes

Nothing here calls what `GeometricOptimizers` 0.5.0 removed — `apply_toNT`, the pirated `ParameterHandling` methods on Base types, or `NeuralNetworkParametersExt`, whose contents are `src/parameter_protocol.jl` there now. The three uses of `NetworkParameters` in `src/nvi/utilities.jl` are unparameterized annotations, so the container's new element-type parameter is invisible here.

## Verification

Full suite passes — 1982 passed, 2 broken — resolved from the registry alone at `GeometricOptimizers` 0.5.0, `NeuralNetworkParameters` 0.2.1, `SymbolicNeuralNetworks` 0.7.0 and `AbstractNeuralNetworks` 0.7.1. No `[sources]` entry and nothing dev'd; the `pre-push` hook ran it again before this branch was pushed.

## Note on the base

This branches from `main`, not from `abstractneuralnetworks-0.7`. That branch's only difference from `main` is narrowing `SimpleSolvers` back to `"0.12.1"`, which would undo the CompatHelper bump to `"0.12.1, 0.13"` that `main` already carries, so it contributes nothing here.

If CI fails in `Pkg.resolve` rather than in a test, check whether the runner's cached General clone is stale before reading the compat bounds — `registry failed to rebase on origin/master` has been the cause repeatedly today, in CI and locally.


michakraus (commented):
## Review

The three bounds are right and they do have to move together — `GeometricOptimizers` 0.5.0 needs the 0.2 container, `SymbolicNeuralNetworks` 0.6.0 forbids it, and 0.7.0 is the release that permits it. I checked the removals claim too: `apply_toNT`, the pirated `ParameterHandling` methods and `NeuralNetworkParametersExt` have no callers anywhere in `src/`, `test/`, `benchmark/`, `docs/` or `scripts/`.

But **"no code changes" is only true of what those releases removed.** Two things they *add* are load-bearing here, and there are three defects in the bounds themselves. Follow-up commits on this branch address all of them.

---

### Major

**1. No version bump — this cannot be registered.** `0.4.0` is already in General (`166fb4365da7ab88f5c948bc9545e69a14872081`). A compat change on top of a registered version needs a new one. Bumped to `0.4.1`: compat-only, nothing in the exported surface moves.

**2. `AbstractNeuralNetworks = "0.7"` admits a version that can never be used here.** 0.7.0 pins `NeuralNetworkParameters = "0.1"`, so it cannot coexist with this PR's `"0.2"`. The resolver works that out on its own and picks 0.7.1, so nothing breaks — but a bound is a statement about what the package supports, and this one is making a false one. `SymbolicNeuralNetworks` 0.7 tightened to `"0.7.1"` for exactly this reason and says so in a comment. Tightened here to match.

**3. `NeuralNetworkParameters = "0.2"` is not the bound this package's test suite needs.** Not about what loads — every name used here is in 0.2.0. It is about what `test/quality/inference_and_allocations.jl` measures; see the next item. Tightened to `"0.2.1"`, with the reasoning in the file.

**4. The Julia 1.10 allocation ceiling this PR inherits is now obsolete — and it is the thing these releases were supposed to close.** `test/quality/inference_and_allocations.jl` carries a 1.10-only ceiling of 42 000 for `ShallowNet` and `ShallowNetReversible`, with a comment ending *"remove this override when that closes."* SNN 0.7.0's changelog says it closed: the cause was a `map` over a closure that 1.10 does not always elide, on the walk that splits a generated function's flat result back into the nesting of the parameters, and the fix has two independent halves — SNN 0.7.0 for the batched walk, NNP 0.2.1 for the un-batched one. This package's symbolic `residual!` calls `DQDθ` on a length-one `Vector`, so it takes the un-batched path, which is why item 3 is a `0.2.1` bound and not a `0.2` one.

That is the whole reason SNN#55 was reported from #86 in the first place. Measured on Julia 1.10.11, macOS aarch64, with nothing but those two versions varied:

| | SNN 0.6.0 / NNP 0.1.1 | SNN 0.7.0 / NNP 0.2.1 |
|---|---|---|
| `ShallowNet` | 28 096 | **15 168** |
| `ShallowNetReversible` | 28 096 | **15 168** |
| `ShallowNetAutodiff` | 51 584 | 51 584 |
| `ShallowNetAutodiffReversible` | 51 584 | 51 584 |

Only the two symbolic rows move, as they did going the other way. The override is removed, the tight 17 000 applies on every version again, and the *Open Issues → Upstream* entry in the CHANGELOG is struck as closed.

**5. `optimizer_params`/`network_params` can go, and should.** Both are `@generated` helpers that flatten the layer nesting into `L1_W`-style keys and back, because `Optimizer` takes a *flat* `NamedTuple` where network parameters are one level deeper. That shape was never what the optimizer works in: handed a `NamedTuple`, `GeometricOptimizers` flattens it itself inside the `Gradient` it builds (`GradientAutodiff(F, ::NamedTuple)` in `src/optimizers/named_tuple_wrapper.jl`), so the flat `NamedTuple` bought a **second, nested unflattening on every function and gradient evaluation of every epoch**.

`NeuralNetworkParameters` 0.2's `flatten`/`unflatten`/`unflatten!` replace both. What makes that possible *now* and not before is precisely what GO 0.5.0 changed: a `ParameterLayout` is a **value**. The `ParameterHandling.flatten` that was the alternative returned a chain of nested closures, one per level of the tree, which is not type stable — and that instability is exactly why these two helpers were `@generated` in the first place. Their own docstrings say so.

The three call sites (`ShallowNet`'s `TrainingMethod` seed, `DenseNet`'s `TrainingMethod` and `LSGD` seeds) lose the aliasing the flat view gave them, so each gains an explicit `unflatten!` where the network's arrays are read back — once after `solve!` for `ShallowNet`, once per epoch for the two hand-rolled `DenseNet` loops, which read the loss off `PNN.params` inside the loop. `test/unit/optimizer_params_unit.jl` becomes `test/unit/parameter_flattening_unit.jl` and pins the contract this package now depends on, including that `unflatten` preserves a `Dual` element type — without which every gradient in those loops would silently come out zero.

### Checked and deliberately *not* done

- **Handing a `NetworkParameters` straight to `Optimizer`.** GO 0.5 makes `NetworkParameters{T}` a member of `OptimizerSolution{T}`, which looks like the invitation. It is not: GO 0.5's changelog under *Known issues* records that the swap itself is not in the release — every elementwise primitive in `named_tuple_wrapper.jl` (`_zero`, `_copy`, `_similar`, `_copyto!`, `_mul!`, `_rac!`, `_div!`, `_square!`, `rgrad`) still dispatches on `ArrayNamedTuple`, and `default_gradient` has no container method. A container "gets further in before failing rather than being turned away at the door." Worth recording so nobody re-investigates.

- **Replacing `flatten_params!` with `NeuralNetworkParameters.flatten!`.** Same walk, but allocation-free only when handed a stored `ParameterLayout`, and the four call sites in `DenseNet`'s `components!` flatten a *freshly built* gradient set with nowhere on the cache to keep a layout for it. The `@generated` walk needs none. `_param_arrays`'s docstring now says this.

### Minor

**6. No CHANGELOG entry.** `[Unreleased]` was empty. Added a `[0.4.1]` section.

**7. `benchmark/Project.toml` has an absolute `[sources]` path** — `/Users/mkraus/Datashare/Julia/NonlinearIntegrators`, in since 8819820 — so that environment resolves on exactly one machine. `docs/Project.toml` has always used `".."`. Fixed; pre-existing and unrelated to this PR, picked up on the way past.

### On the PR description

The note about a stale General clone in CI is worth keeping, but "check whether the runner's cached General clone is stale **before** reading the compat bounds" now cuts the other way for item 2: `AbstractNeuralNetworks = "0.7"` resolving fine locally is exactly what a correct bound and a *too wide* one look like from the outside.

---

### Verification

- Full suite on **1.13.0-rc2**: 1981 passed, 2 broken. The count is one below this branch's 1982 because the replacement test file has one fewer `@test`, not because anything stopped being checked.
- Full suite on **1.10.11**: passes with the 17 000 ceiling in force — the run the removed override existed for.
- **No runtime cost from the refactor**: 11m56s against 11m34s for this branch unmodified, same machine, both quiet. (An earlier 46m reading on the modified tree was mine — it overlapped a full second-environment precompile — and is not a real signal.)
- The allocation figures above come from a probe env holding the rest of the stack fixed. Worth knowing if you reproduce it: `[sources]` is Julia 1.11+, so on 1.10 a probe env with a `[sources]` path entry silently resolves the *registered* `NonlinearIntegrators` instead of the working tree. That is how the 28 096 column got measured by accident before it was measured on purpose.

PR #96, opened 2026-08-29
Take the de-piracy wave: NeuralNetworkParameters 0.3 and the five bounds behind it

michakraus:
Six compat bounds move together, and none of them reaches code here.

## One constraint, not six

`NeuralNetworkParameters` 0.3.0 removes `ParameterSet`, the `Union{NetworkParameters, NamedTuple}` this ecosystem dispatched on, because a method written on it was a method on `Base.NamedTuple`. Every other bound follows from that one:

| | was | is |
|---|---|---|
| `NeuralNetworkParameters` | `0.2.2` | **`0.3`** |
| `AbstractNeuralNetworks` | `0.7.1` | **`0.8`** |
| `SymbolicNeuralNetworks` | `0.7` | **`0.8`** |
| `GeometricOptimizers` | `0.6` | **`0.7`** |
| `SimpleSolvers` | `0.12.1, 0.13` | **`0.13.2`** |
| `GeometricBase` | `0.14.8` | **`0.14.9`** |

The middle three are the first releases of their packages that permit the 0.3 container; `GeometricOptimizers` 0.7.0 requires `SimpleSolvers` 0.13.2; and `NeuralNetworkParameters` 0.3.0 takes `GeometricBase` as a hard dependency at 0.14.9, for `L2norm` over a parameter set. Any one of them alone is unsatisfiable.

## Nothing here changes

`src/` and `test/` name neither `ParameterSet` nor `EquationSet` nor `ArrayNamedTuple`. The call sites reach only `NetworkParameters`, `params`, `flatten`, `unflatten` and `unflatten!`, none of which changed.

`GeometricOptimizers` 0.7.0 takes a whole set of parameters **only** as a `NetworkParameters` and turns a bare `NamedTuple` away at `Optimizer`. That is the breaking half of the wave and it costs this package nothing, for the reason 0.4.2's note already gave: the `ShallowNet` initialisation flattens to a plain `Vector` before `Optimizer` sees it (`src/nvi/shallownet.jl`), so no parameter set is ever handed over whole.

## The `residual!` budgets moved, downwards

Re-measured rather than carried over:

| | 0.4.2 | 0.4.3 |
|---|---|---|
| `ShallowNet` | 11 424 | 11 424 |
| `ShallowNetReversible` | 11 424 | 11 424 |
| `ShallowNetAutodiff` | 54 656 | **52 096** |
| `ShallowNetAutodiffReversible` | 54 656 | **52 096** |

Worth stating which way, because 0.4.2 moved these same two rows *up* by 3 072 and left the symbolic pair alone. This moves them back down by 2 560 and again leaves the symbolic pair alone — so the autodiff path, which reaches its derivatives through `ForwardDiff` rather than a generated kernel, is the one that feels these releases, and the generated one is not. The ceilings (17 000 and 78 000) hold with room, so none of them moves.

## Also

The changelog gains the **0.4.2 entry it was released without**, reconstructed from that release's own commit and marked as written after the fact.

## Version

**0.4.3.** Compat only; the exported surface is unchanged.

## Verified

Full suite passes — 1981 passed, 2 broken, unchanged. The pre-push hook's `Pkg.test` also ran clean, which it could not have before these releases were registered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,24 +12,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The de-piracy wave, taken as a set of compat bounds. Nothing in this package's exported surface
 changes, and neither does any of its code.
 
-### Fixed
-
-- **`test/unit/network_integrators_unit.jl` was a lottery, and it had been one all along.** Every
-  case there builds a network with random weights, and the OGA/Newton path on those weights is
-  ill-conditioned by nature — `Float32` `ShallowNet` under `OGA1dStable` sits close enough to a
-  singular Gram matrix that whether it *is* singular depends on which weights were drawn. Nothing
-  seeded them: the only `Random.seed!` in the suite is in `dispatch_variants_unit.jl`, which
-  `runtests.jl` includes *after* this file, so the draw depended on how many times `rand` had been
-  called by whatever ran earlier — which differs by platform and by Julia version.
-
-  It surfaced here because this release moved dependency versions, but it is not caused by them.
-  Run against `main`'s **unchanged** bounds on the same runners on the same day, the same test fails
-  the same way with `SingularException(13)`: 5 of 13 jobs there, 8 of 13 here, with identical
-  resolved versions of everything either branch touches, while 24 consecutive local runs on macOS ARM
-  passed under both. `main` being green on 2026-08-26 was a draw, not a state.
-
-  The file seeds itself now, before the cross product and independently of the include order.
-
 ### Changed
 
 - **`NeuralNetworkParameters` 0.2.2 → 0.3, `AbstractNeuralNetworks` 0.7.1 → 0.8,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The de-piracy wave, taken as a set of compat bounds. Nothing in this package's exported surface
 changes, and neither does any of its code.
 
+### Fixed
+
+- **`test/unit/network_integrators_unit.jl` was a lottery, and it had been one all along.** Every
+  case there builds a network with random weights, and the OGA/Newton path on those weights is
+  ill-conditioned by nature — `Float32` `ShallowNet` under `OGA1dStable` sits close enough to a
+  singular Gram matrix that whether it *is* singular depends on which weights were drawn. Nothing
+  seeded them: the only `Random.seed!` in the suite is in `dispatch_variants_unit.jl`, which
+  `runtests.jl` includes *after* this file, so the draw depended on how many times `rand` had been
+  called by whatever ran earlier — which differs by platform and by Julia version.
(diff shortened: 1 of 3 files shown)


michakraus:
## The A/B is in: this failure is not from the bump

[#97](https://github.com/JuliaGNI/NonlinearIntegrators.jl/pull/97) runs `main`'s **unchanged** dependency bounds on today's runners. It fails the same way:

```
ShallowNet OGA1dStable × IntegratorExtrapolation (Float32): Error During Test at test/unit/network_integrators_unit.jl:32
  Got exception outside of a @test
  SingularException(13)
```

Same test, same exception, same `Float32` `ShallowNet` case. Counts differ — 8 of 13 jobs here, 5 of 13 there — which is what nondeterminism looks like, not what a version-dependent regression looks like.

What was ruled out along the way:

- **Not the stale registry.** The first three failures were `Some registries failed to update`; the current run resolves everything correctly.
- **Not a version skew between local and CI.** Both resolve to identical versions of all six changed packages plus `ForwardDiff`, `GeometricBase` and `GeometricIntegratorsBase`.
- **Not reproducible on macOS ARM.** The exact failing case, 12 runs under the new bounds and 12 under main's, in separate environments: **0/12 failed both ways.**

The cause is that `test/testsetup.jl` seeds nothing — there is no `Random.seed!` anywhere in the suite — so an ill-conditioned `Float32` Newton solve is drawn afresh each run and flips on BLAS and CPU differences. `main`'s green run on 2026-08-26 was a lucky draw, not a passing state.

Fixing that is what now stands between this PR and a green CI, so it is going in here rather than being left for later.

michakraus:
## Correction: the seed was the wrong diagnosis, and is reverted

I attributed the failure to an unseeded RNG draw. That was wrong, and seeding did not fix it — same test, same two cases, same `SingularException(13)`, with the line number moving 32 → 52 confirming the seed was live.

**It cannot be the draw.** A fixed seed produces identical weights on every platform, and yet on the *same commit* `1.11`- and `1.12`-macOS pass while `1.13`- and nightly-macOS fail. The variable is the linear algebra, not the weights.

## What it actually is

The Gram matrix for `Float32` `ShallowNet` under `OGA1dStable` is **numerically rank-deficient**, and whether LAPACK's LU declares it exactly singular differs by BLAS build and version. `SingularException(13)` is a pivot at index 13 being exactly zero — a threshold being crossed, not a wrong answer being computed.

This file's own header records the same symptom once before, for `OGA1dNormalEquations` under Hermite, where it turned out to be `network_labels` left at zero rather than the κ(Φ)² conditioning it looked like. Worth checking whether this is the same shape of thing.

## It is not from this branch

[#97](https://github.com/JuliaGNI/NonlinearIntegrators.jl/pull/97) runs `main`'s **unchanged** bounds on the same runners and fails identically — 5 of 13 jobs there, 8 of 13 here, with identical resolved versions of everything either branch touches. `main` last went green on 2026-08-26, before the runner images moved.

The seed is reverted, so this PR is a pure bounds bump again. It is ready apart from a red CI that it did not cause.

